PR #66, opened 2023-01-07
feat: enable case-insensitive search

lordgiotto:
Hi,
just a tiny PR to enable case-insensitive search :)

I hope it looks good to you and it's useful.


godly-devotion:
Looks good, thanks!

PR #142, opened 2023-02-06
Token Counting Changes

CarterLombardi:
Tokens are now calculated using BPETokenizer from the StableDiffusion library.
Added a live counter in the prompt entry text box showing how many tokens you're currently using.
Live counter turns yellow to signal a warning when too many tokens are used.


godly-devotion:
Could you revert the changes on `es.lproj/Localizable.strings`?

PR #160, opened 2023-02-18
Close QuickLook if called for the currently open image

vzsg:
This matches the behavior of QL in other apps, where you can open and close it by repeatedly pressing space.

### DIFF
--- a/Mochi Diffusion/Support/ImageController.swift
+++ b/Mochi Diffusion/Support/ImageController.swift
@@ -32,6 +32,7 @@ final class ImageController: ObservableObject {
 
     @Published
     var quicklookURL: URL?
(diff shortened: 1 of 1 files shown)


godly-devotion:
Interesting, I wonder why setting `quicklookId` works even though it seems its not being used.

godly-devotion:
Never mind I see whats happening.

godly-devotion:
Hmmm it looks that this introduced some regressions.
- Removing another image while Quick Look is open causes Quick Look to close
- If an image is open with Quick Look and is upscaled, the Quick Look view doesn't show the upscaled image but rather closes

vzsg:
Crap... at least both seem easy to reproduce. I'll look into a fix later today.

PR #163, opened 2023-02-19
Implement and improve search functionality

vzsg:
I'm putting this up as a draft PR for transparency and potential feedback.

---

This is my attempt at getting the search functionality to work. It works reasonably well, until the current image is deleted: then the selection moves to the next image of the full set, not of what's currently visible.

Selection state is stored in SDImage, but is updated by ImageStore, while some actions to do it are in ImageController; while only GalleryView knows about the filtered list (this is new). This all needs to be consolidated.


vzsg:
I was thinking of
- removing the SDImage.isSelected property, making the selected image ID "external"
- moving the filtering logic out of GalleryView, into or somewhere closer to ImageController
- removing all traces of selection from ImageStore, moving closer to ImageController again
- doing something arcane to split up ImageController (I'm still daydreaming about this part)

godly-devotion:
I think having the selection state within SDImage is a better choice than being external for several reasons. If selection state is external, any time the order or SDImage is updated, the selection index will need to be carefully updated in all parts. However if the property is already part of the model, potential state mismatch issues can be avoided. Looking for the selected item(s) also becomes as easy as filtering for the `isSelected` property. Also handling multi select becomes easy as well.

vzsg:
The selected index is not actually used or important (nor will be possible if multiple selection becomes a thing). In my second commit, I've already moved the selection into the ImageStore, and it did not have any negative effect on neither the code, nor the performance. If anything, there are less loops to process on change.

vzsg:
Sidenote: and if the O(N) lookups would become a problem in ImageStore, we could drop in Apple's [OrderedDictionary](https://github.com/apple/swift-collections/blob/main/Documentation/OrderedDictionary.md) to back it instead of a plain Array.

Mochi Diffusion/Views/AppView.swift
```diff
@@ -25,12 +25,15 @@ struct AppView: View {
                     .frame(maxWidth: 340)
             }
         }
-        .searchable(text: $galleryConfig.searchText, prompt: "Search")
+        .searchable(text: Binding(get: { store.searchText }, set: { store.searchText = $0 }), prompt: "Search")
```
godly-devotion, line 28: Is there a reason for the manual binding?

vzsg: Good call, I meant that as a temporary measure to get things to compile. I'll clean it up.

vzsg: Yeah, my bad. It slipped my mind that EnvironmentObjects can be projected too. 😓

godly-devotion:
> The selected _index_ was not stored even before my changes, and it's only calculated transiently inside ImageController to move the selection back-and-forth. (Such a selected index will not be possible if multiple selection becomes a thing).
> 
> In my second commit, I've already moved the selection into the ImageStore, and it did not have any negative effect on neither the code, nor the performance. If anything, there are less loops to process on change and much less state that can get desynchronized. State is hell.

Sorry I meant selected ids.
For example, to decide if the selected border style should be applied to a gallery item, we can just use the `isSelected` property to easier figure that out. For ids, we will need to use the selectedIds property and match the id with each item.
Also when doing something like remove, if the item was `isSelected == true`, nothing more needs to be done, but with selectedIds, we would need to update it. It just feels easier keeping it within each model. Though if I am missing something, feel free to explain it to me.

Mochi Diffusion/Model/ImageStore.swift
```diff
@@ -45,62 +54,62 @@ class ImageStore: ObservableObject {
     func remove(_ id: SDImage.ID) {
         withAnimation {
             guard let index = index(for: id) else { return }
-            images.remove(at: index)
+            allImages.remove(at: index)
```
godly-devotion, line 57: Shouldn't we check here if the item was part of the `selectedId(s)` here and remove the SDImage from it if it is?

vzsg: True, this is work in progress.
The selection state will have to be aligned with the filter too, so the UX is not weird.

godly-devotion:
I did some more thinking on how expensive a view refresh would be for multi select state and put my observations in the table. Assuming 20 items with 3 items selected with indices 12, 13, 14 (start index == 0). View is recalculated every time `images` changes. Loop for `selectedIds` increases as # of items selected increases.
| ImageStore changes                          | `sdi.isSelected`    | `selectedIds.contains(sdi.id)`                               |
| ------------------------------------------- | ------------------- | ------------------------------------------------------------ |
| Setting selection                           | 20 loops            | None                                                         |
| Filter selected items                       | 20 loops            | 14 loops (highest index)                                     |

| GalleryView view updates                    | `sdi.isSelected`    | `selectedIds.contains(sdi.id)`                               |
| ------------------------------------------- | ------------------- | ------------------------------------------------------------ |
| Drawing selection border on item add/remove | 20 loops (per item) | 20 (per item) x up to 3 (selectedIds.count) = up to 60 loops |

vzsg:
What exactly are loops here?

godly-devotion:
Either for loops or something similar. Basically the amount of iterations of `images` array required for whatever action/change occurs.

vzsg:
I think you're overestimating the steps needed with the Set, perhaps even in the sdi.isSelected case too. There is no reason to iterate through images to decide whether an item view needs a border or not, since in the first case the isSelected state is directly available, and in the second case, the ID is directly available with which the O(1) lookup in the selectedIds Set can be performed. Same number of operations, regardless of the number of selected items.

In fact, since the images array does not change anymore, including the SDImages inside, I think there's even less reason for SwiftUI to re-evaluate the whole array for the ForEach. (This might be why the blinking divider glitch is gone.)

But do correct me if I misunderstood the operations.

godly-devotion:
Perhaps, this was my thinking process.
Untitled

godly-devotion:
Yeah you're right, I think using `selectedIds` may be better and more intuitive. If needed, we can always use a HashTable for `O(1)` lookups (could be overkill) but for now arrays should be fine.

godly-devotion:
For the focus of this PR, we won’t worry about multi select.

vzsg:
I'm done with rebuilding the search and selection logic. I added something new in the last commit: if there's no selected image, select next will select the first image, select previous will select the last image, and moving the selection will wrap around. For me, it feels rather natural, and it helps a lot if the selected image disappears in a search.

godly-devotion:
Just tried it out, works like a charm 👍

PR #164, opened 2023-02-21
Fix QuickLook regressions after upscale and removal

vzsg:
The quicklookId is now the single source of truth, and quicklookURL is derived from it, instead of mutating both in parallel. This fixes both issues reported in the comments of #160.


godly-devotion:
That was quick. Thanks!

godly-devotion:
Oh no, if you open an image in Quick Look, close it, and then remove an image, Quick Look shows up again 😂

vzsg:
Great... I wonder why it didn't immediately happen when I was testing it 😦 I'll figure something out.

PR #180, opened 2023-02-27
Consider symlinks when scanning for model subdirs

surjikal:
With this change, I can use symlinks to follow MochiDiffusion's desired model folder structure and not mess with my model repos.

This is more of a feature request and quick hack to make it work for myself.
I am not familiar with swift.


godly-devotion:
What's wrong with setting the custom model folder? Do you have models under multiple subfolders?

surjikal:
Yeah pretty much.

I cloned a bunch of models, but since they're not in a flat structure, I had to copy the data in a way that mochi can read them. But then I lose the ability to do a `git pull` to update those models. 

For example, this model here:
https://huggingface.co/apple/coreml-stable-diffusion-2-1-base/tree/main

I set up symlinks like this:
- /coreml-stable-diffusion-2-1-base/original/compiled -> /mochi/models/coreml-sd-2-1-original-compiled
- /coreml-stable-diffusion-2-1-base/split_einsum/compiled -> /mochi/models/coreml-sd-2-1-split-ensum-compiled

TBH I'm new to setting up stable diffusion, maybe there's no point to doing `git pull` on models or whatever. But beyond my use case maybe supporting symlinks is useful to someone else.

godly-devotion:
Just FYI cloning the model directory is a bad bad idea. Git keeps history of all changes so your repo size will grow really large if there happens to be an model update.

KengChiChang:
I'd like this change too since I store the model weights on an external NAS drive so that I can save more disks space on my machine.

1ucas:
I was about to open a PR on this too. 
Symlinks would be very useful.

Answering the question:
`What's wrong with setting the custom model folder? Do you have models under multiple subfolders?`
Yes.
Let's consider the following folder structure:

StableDifussion (root)
 - server/clients
   - **(any native or web-ui style stable diffusion clients here)**
 - models
     - sd15-models
       - **symlink of all sub-folders containing Stable Diffusion 1.5 models inside each style folder below**
   - realistic
     - deliberate
     - realistic-vision
     - cyber
   - anime
     - dreamshaper
     - anything
   - landscapes
     - colorful

With symlink support I can choose the `sd15-models` folder as the _custom model folder_ and have all models referenced in the App without losing the original organization nor having to duplicate all models to the `sd15-models` folder.

jrittvo:
The recursive subfolder issue was what I was asking about. But you are right.  That can and does trip people up with or without supporting symlinks, so no new damage done there.  Thanks for your response.

PR #170, opened 2023-02-23
Fix crash when selecting an invalid model

vzsg:
Fixes #165.

And improves the data race that happened between the compute unit popover and the model loading by asking first, then doing the load.

It's working locally (both ARM and Rosetta), but there is still one issue I haven't figured out: when I dismiss the popover from the controller with `presentingComputeUnitSelection = false`, the apps spits out 8 lines of SwiftUI warnings:

```
[SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
```

I know by experiments that it's not timing-related, since a DispatchQueue.main.asyncAfter does not fix it.

---

If you have an idea how to evade it, or if we could just ignore it, please let me know.

PR #209, opened 2023-03-29
Detect model type automatically

### LINKED ISSUE #193: Subfolder in folder model causes CTD
### Running Latest Version

- [X] I am running the latest version

### Processor

M1 (or later)

### Intel processor

_No response_

### Memory

16GB

### What happened?

creatred a "noi2i" folder inside the model folder to organize things. "noi2i" appears in the model list, selecting this crashes mochi diffusion.

### Crash Log

_No response_

vzsg:
I had a flash of inspiration yesterday, and found a way to distinguish original/split-einsum models without introducing custom JSONs or anything extra into/around the models.

This renders the unit selection popover unnecessary too, so I dropped that and introduced a new setting as the "compute unit preference", replacing the previous, directly set MLComputeUnits. The default for everyone is Auto, but advanced users can force a specific unit in Settings/Image as before. Help labels are updated to match.

Fixes #193 and #170 (since the offending popover and race condition is gone).

(And I found a fun race condition when selecting a new model and pressing Generate twice causing two images to be rendered at the same time, but there's no open issue for it.)


godly-devotion:
Very nice! I'll pull the changes and take a look today or tomorrow.

PR #214, opened 2023-04-02
Support Homebrew installation

missuo:
**Support Homebrew installation.**

### DIFF
--- a/Casks/mochi-diffusion.rb
+++ b/Casks/mochi-diffusion.rb
@@ -1,0 +1,13 @@
+cask "mochi-diffusion" do
+  version "3.0"
+  url "https://github.com/godly-devotion/MochiDiffusion/releases/download/v#{version}/MochiDiffusion_v#{version}.dmg"
+  sha256 "0d91f07fdfb8780cbeae01cc439b2dc613e2dc3062699ffa38b504bd541d7d67"
+
+  name "MochiDiffusion"
+  desc "Run Stable Diffusion on Mac natively"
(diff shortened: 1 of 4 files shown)


godly-devotion:
Does the cask file need to be part of the repo? I personally haven't seen this in other repos.

missuo:
Required. The `.rb` file in the casks directory contains the release information. Brew will find the corresponding release based on this file and install it automatically. 

However, when a new version is released, you may need to manually change the version and sha256 of the `.rb` file, but you can also leave it to GitHub Actions to update it if you want to automate it.

vzsg:
It might also be possible to get included in the [homebrew-casks repo](https://github.com/Homebrew/homebrew-cask/pulls), so this one doesn't need to be aware.

missuo:
> It might also be possible to get included in the [homebrew-casks repo](https://github.com/Homebrew/homebrew-cask/pulls), so this one doesn't need to be aware.

This is of course the better way, but each version update initiates a pr to the `homebrew-casks` repo until it is merged.

godly-devotion:
Honestly since its already available in [homebrew](https://formulae.brew.sh/cask/mochi-diffusion) and Mochi Diffusion has an built-in update checker/updater already, I don't think having the cask version trail behind a few days is terrible.
There's also the fact that I don't want to manage the cask.

Mrered:
Third-party support

```sh
brew tap ivaquero/chinese
brew update
brew install mochi-diffusion
```

PR #232, opened 2023-04-14
Added animation feedback when converting to high resolution

jinhongw:
I have added an animation effect feedback that appears when the 'Convert to High Resolution' button is clicked. This allows users to easily understand that the picture is being converted to high resolution. 

Additionally, I have changed the warning text color from systemYellow to systemOrange due to clarity issues caused by the previous prompt word being too long.


Mochi Diffusion/Views/SidebarControls/PromptView.swift
godly-devotion, line 1: This seems to have been accidentally committed, could this be reverted?

Mochi Diffusion/Views/GalleryItemView.swift
```diff
@@ -7,17 +7,138 @@
...
+                Rectangle()
+                    .foregroundColor(.black)
+                    .blur(radius: 50)
+                    .opacity(0.3)
+
+                /// Left Top WandStars
+                ForEach(0..<2) { _ in
+                    let x = CGFloat.random(in: 0...(geometry.size.width / 2))
+                    let y = CGFloat.random(in: 0...(geometry.size.height / 2))
+                    WandStar(size: CGFloat(Int.random(in: sizeRange)))
```
godly-devotion, line 42: Since only the `x` and `y` coordinates are different, I think we should be able to consolidate the `WandStar` view rather than verbosely repeating it.

jinhongw: I modified it, now `WandStar` view no longer has duplicate code.

godly-devotion:
Looks good, thanks!

PR #242, opened 2023-04-26
Allow custom image sizes for img2img

### LINKED ISSUE #208: Possible Fix For image2image 512x768 and 768x512
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

Someone appears to have found that the width and height parameters are transposed between the t2i and i2i parts of the Automatic1111 pipeline on MacOS.  Perhaps this same issue is the root of why Mochi has only been able to implement i2i at 512x512.  Please look at this thread section and decide if this is worth investigating.  https://github.com/apple/ml-stable-diffusion/issues/143#issuecomment-1487612611

### Why do you think this should be added?

This was something you hoped to include in the i2i implementation, I believe.

-------------------------------------------------------------------

Screenshot

vzsg:
Closes #208.
With the underlying issue gone, there's no need to limit the users to 512x512.

### DIFF
--- a/.github/workflows/analyze-build.yml
+++ b/.github/workflows/analyze-build.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-13
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


vzsg:
That build failure is quite unfortunate. The ml-stable-diffusion repo needs Swift 5.8 (Xcode 14.3) now, but the workflow only installs Xcode 14.2.

PR #245, opened 2023-04-27
Add ControlNet

stuartjmoore:
Adds the ability to use a ControlNet model with an accompanying image.

## Screenshots
Screenshot 2023-06-03 at 3 54 55 PM
Screenshot 2023-06-03 at 3 55 06 PM
Screenshot 2023-06-03 at 3 55 13 PM
Screenshot 2023-06-03 at 3 58 36 PM

### DIFF
--- a/Mochi Diffusion/Model/SDModel.swift
+++ b/Mochi Diffusion/Model/SDModel.swift
@@ -19,7 +19,7 @@ struct SDModel: Identifiable, Hashable {
 
     var id: URL { url }
 
-    init?(url: URL, name: String) {
+    init?(url: URL, name: String, controlNet: [String]) {
         guard let attention = identifyAttentionType(url) else {
             return nil
         }
@@ -27,19 +27,8 @@ struct SDModel: Identifiable, Hashable {
         self.url = url
         self.name = name
         self.attention = attention
-        self.controlNet = controlNets(url)
+        self.controlNet = controlNet
     }
-}
-
-private func controlNets(_ url: URL) -> [String] {
-    let controlNetSymLinkPath = url.appending(component: "controlnet").path(percentEncoded: false)
-
-    guard FileManager.default.fileExists(atPath: controlNetSymLinkPath),
-        let contentsOfControlNet = try? FileManager.default.contentsOfDirectory(atPath: controlNetSymLinkPath) else {
-        return []
-    }
-
-    return contentsOfControlNet.filter { !$0.hasPrefix(".") }.map { $0.replacing(".mlmodelc", with: "") }
 }
 
 private func identifyAttentionType(_ url: URL) -> SDModelAttentionType? {
--- a/Mochi Diffusion/Model/SDModel.swift
+++ b/Mochi Diffusion/Model/SDModel.swift
@@ -32,10 +32,10 @@ struct SDModel: Identifiable, Hashable {
 }
 
 private func controlNets(_ url: URL) -> [String] {
-    let controlNetSymLink = url.appending(component: "controlnet")
+    let controlNetSymLinkPath = url.appending(component: "controlnet").path(percentEncoded: false)
 
-    guard FileManager.default.fileExists(atPath: controlNetSymLink.path()),
-        let contentsOfControlNet = try? FileManager.default.contentsOfDirectory(atPath: controlNetSymLink.path()) else {
+    guard FileManager.default.fileExists(atPath: controlNetSymLinkPath),
+        let contentsOfControlNet = try? FileManager.default.contentsOfDirectory(atPath: controlNetSymLinkPath) else {
         return []
     }
 
--- a/Mochi Diffusion/Model/SDModel.swift
+++ b/Mochi Diffusion/Model/SDModel.swift
@@ -35,7 +35,7 @@ private func controlNets(_ url: URL) -> [String] {
     let controlNetSymLink = url.appending(component: "controlnet")
 
     guard FileManager.default.fileExists(atPath: controlNetSymLink.path()),
-          let contentsOfControlNet = try? FileManager.default.contentsOfDirectory(atPath: controlNetSymLink.path()) else {
+        let contentsOfControlNet = try? FileManager.default.contentsOfDirectory(atPath: controlNetSymLink.path()) else {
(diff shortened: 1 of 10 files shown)


Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -109,14 +110,40 @@ class ImageGenerator: ObservableObject {
...
+                    let controlledUnetMetadataPath = url.appending(components: "ControlledUnet.mlmodelc", "metadata.json").path(percentEncoded: false)
+                    let controlNetSymLinkPath = url.appending(component: "controlnet").path(percentEncoded: false)
+
+                    if fm.fileExists(atPath: controlledUnetMetadataPath), !fm.fileExists(atPath: controlNetSymLinkPath) {
+                        try? fm.createSymbolicLink(atPath: controlNetSymLinkPath, withDestinationPath: finalControlNetDirURL.path(percentEncoded: false))
+                    }
+
+                    return url
+                }
                 .compactMap { SDModel(url: $0, name: $0.lastPathComponent) }
```
stuartjmoore, line 147: I didn’t want to pass the ControlNetDir in the model struct, so I’m preemptively creating the symlink if needed.

Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -109,14 +110,40 @@ class ImageGenerator: ObservableObject {
...
+            finalControlNetDirURL = fm.homeDirectoryForCurrentUser
+            finalControlNetDirURL.append(path: "MochiDiffusion/controlnet/", directoryHint: .isDirectory)
+        } else {
+            finalControlNetDirURL = URL(fileURLWithPath: controlNetDir, isDirectory: true)
+        }
+
+        if !fm.fileExists(atPath: finalControlNetDirURL.path(percentEncoded: false)) {
+            print("Creating ControlNet directory at: \"\(finalControlNetDirURL.path(percentEncoded: false))\"")
+            try? fm.createDirectory(at: finalControlNetDirURL, withIntermediateDirectories: true)
+        }
```
stuartjmoore, line 131: I feel like this (and the copied-from `modelDir` lines above) could move up one level to before where `getModels` is called. That would save returning the URLs in the too-large tuple and focus this method a bit more.

Mochi Diffusion/Model/SDModel.swift
```diff
@@ -26,18 +27,39 @@ struct SDModel: Identifiable, Hashable {
...
+private func controlNets(_ url: URL) -> [String] {
+    let controlNetSymLinkPath = url.appending(component: "controlnet").path(percentEncoded: false)
+
+    guard FileManager.default.fileExists(atPath: controlNetSymLinkPath),
+        let contentsOfControlNet = try? FileManager.default.contentsOfDirectory(atPath: controlNetSymLinkPath) else {
+        return []
+    }
+
+    return contentsOfControlNet.filter { !$0.hasPrefix(".") }.map { $0.replacing(".mlmodelc", with: "") }
+}
```
stuartjmoore, line 43: There’s no need to do this every single time for every model; I’ll move this up a level and run it once when loading.

godly-devotion:
Regarding your question about supporting multiple ControlNets, I currently don't see much of a use case for it for now. If thats the case, I think we can simplify the UI to this:
image

jrittvo:
This would seem logical to me, but might be confusing to other users . . . 

If you go with the single CN, could the present "STARTING IMAGE" UI module be used for both i2i and CN?  The first choice in the drop down could always be image2image.  If the user picks a CN instead, the strength slider would become inactive.  Using i2i and a CN together may not even be possible in Apple's current implementations, and you could always get the same result doing one generation after the other.  I'd prefer the simpler look of the interface that way.

godly-devotion:
> This would seem logical to me, but might be confusing to other users . . .
> 
> If you go with the single CN, could the present "STARTING IMAGE" UI module be used for both i2i and CN? The first choice in the drop down could always be image2image. If the user picks a CN instead, the strength slider would become inactive. Using i2i and a CN together may not even be possible in Apple's current implementations, and you could always get the same result doing one generation after the other. I'd prefer the simpler look of the interface that way.

That is something we'd have to test with the cli. However if Apple does allow setting both the starting image and ControlNet, we can allow that as well.

jrittvo:
I may have tested that over the weekend when we were trying to see if the i2i starting image would magically act as the input image to the InPaint CN (image plus mask, with prompt to changed masked area).  I saved the CLI prompts I used.  I'll give a look at them in the next day and see if they were the same as the prompt one would use for i2i and a CN in a single command.  I should also try with a simple CN like Canny rather then the 2-input InPaint.  But you can only have a single prompt in the CLI, so I don't know how you could tell i2i to "change the man's hair color from brown to blond" and prompt the Canny with "woman" to apply to the outline mask it constructs. Beyond that, the pipeline did not seem to be able to pass the result of the first operation to be the start of the second operation.  That was a roadblock to multiple CNs as well.

jrittvo:
The results using the same input/prompt/seed of ControlNet alone, image2image alone, and ControlNet and image2image together, are all different.  I want to do one more set and then I'll post the grids.  I'll put them in models at Discord.  I'll leave to someone else to suggest what is happening, and if it is useful.  

The "strength" argument working now.  Based on what I see it doing, I want to go back and test the InPaint CN again with strength functioning.  That'll be for tomorrow.

godly-devotion:
Thanks for the changes! It seems app crashes if ControlNet image is selected without a ControlNet available. Also I couldn't found a way to clear the image.

jrittvo:
I cut and pasted this together yesterday.  It is without the corner X (which I agree is elegant) but the real thing it is trying to suggest is moving both i2i and CN below the set of inputs that always apply, even without i2i or CN.

Mockup

x4080:
waiting for this

theboylzh:
> Regarding your question about supporting multiple ControlNets, I currently don't see much of a use case for it for now

Multiple controlnet is the great usage to control the image. For example, in Automatic webui, I usually use 2-3 controlnet to make the different movements and expressions of consistent characters. So I am looking forward to this feature very much.

PR #256, opened 2023-05-24
Starting Image: Change strength value range

jrittvo:
This PR changes the range of the "STARTING IMAGE" "STRENGTH" slider to 0.645 - 1.000 in 10 increments of 0.035.  This range more closely tracks the effective range of "STRENGTH" changes to the generated image.

This PR also revises the info text for "STARTING IMAGE" to try to more accurately reflect the function of the strength slider in the i2i pipeline.


godly-devotion:
Thanks for the changes! I'm going to revert the message until we find a better way to describe it and to avoid needing to retranslate that section right now.

PR #270, opened 2023-07-04
feat: Added preview for generating images.

hoseins77:
In this pull request, I have implemented the image generation preview feature. Now, users can conveniently visualize the step-by-step process of generating images. Additionally, I have included a toggle functionality to enable or disable the image preview. This enhancement opens up possibilities for future additions, such as a "skip current image" feature and other exciting functionalities.


jrittvo:
I pulled this into my local build.  It's fun to watch!  The toggle is great, very convenient.  Cool feature all around.  Thanks!

hoseins77:
@jrittvo yes its cool. Im thinking to add some functionalities later. If you have something in mind let me know.

Gomeo:
> @jrittvo yes its cool. Im thinking to add some functionalities later. If you have something in mind let me know.

If you have discord, you might consider joining the Mochi Diffusion Discord server. A few devs in there commonly discuss current features and trouble shoot problems. May give you up ideas for functionalities!

https://discord.gg/x2kartzxGv

hoseins77:
@Gomeo I have joined on discord server but i dont have developer role.

jrittvo:
Very little happens in the developer channel. Mostly administrative stuff. Check out all the other channels for discussions on features and improvements. You’ll get automatically added to developer when a PR gets merged.

godly-devotion:
Thanks for the PR, I'm slightly busy this week so I'll probably be able to take a look around Sunday. @hoseins77 I've given you developer role.

hoseins77:
@godly-devotion Thank you

godly-devotion:
I got to take a proper look at it. Its pretty nice to be able to see the process. I have two comments:

1. We should move the "Show preview" option to the Settings window.
2. When the option is checked, the main UI lags really badly. I can't seem to even scroll through the images.

jrittvo:
There is a new PR today at ml-stable-diffusion that makes denoised update images available in the Swift pipeline for use with image preview.  The effect is quit different.  The PR has good examples of without and with noise.  I don't know if the change, if merged, would require any code changes in the PR here.

https://github.com/apple/ml-stable-diffusion/pull/210

ZachNagengast:
Hi, didn't realize there was a PR for this here too! Yea I wasn't sure if changing the behavior of `currentImage` would impact anyone else, since now the noisy latent images won't be accessible. Open to suggestions, I just considered the denoised image to look a little better. 

Also regarding the impact on performance, in this PR https://github.com/huggingface/swift-coreml-diffusers/pull/65 there is now a setting to only compute the `currentImage` every so often, so you still can see it generating but can get a little performance back.

jrittvo:
We will defer to whatever ml-stable diffusion ends up doing.  The MochiDiffusion PR is just to start using the image preview from ml-stable-diffusion in whatever form it is in the Mochi UI.  The noisy image stream is maybe more "interesting", but the denoised image stream is more useful in terms of deciding to interrupt a long inference that doesn't look promising.

jrittvo:
@hoseins77 -  It looks like the PR at Apple ml-stable-diffusion is going to move ahead, with 2 option that you could use in your PR here.  Besides on/off, you can have a "use noisy image" or "use denoised image" option for the preview, and a setting for "update preview every x steps".

jrittvo:
And it just merged.  You're at bat again @hoseins77.

We sure are getting a lot of new toys all of a sudden!

jrittvo:
This PR still works fine with the updated ml-stable-diffusion from the past few days.  It would be nice though to have the option to choose between noised previews and the new denoised previews that can now be accessed.

godly-devotion:
> @hoseins77 - It looks like the PR at Apple ml-stable-diffusion is going to move ahead, with 2 option that you could use in your PR here. Besides on/off, you can have a "use noisy image" or "use denoised image" option for the preview, and a setting for "update preview every x steps".

Nice! I think the best solution would be to use the denoised method similar to that @ZachNagengast implemented, with the gallery & settings UI that @hoseins77 implemented in this PR. I'm open to those changes if @hoseins77 or someone else wants to give it a shot.

jrittvo:
Absent a way for the user to chose "update every X steps" I think that a fixed value has to be hard coded.  I thought that ml-stable-diffusion was going to default it to "no updates", so the stock behavior would be completely off, but when I ran it last night, it was still trying to update faster than my machine could handle.  I suggest fixing it at 1:2 or 1:3 so it doesn't increase generation time for updates that get dropped anyway.  Also agree on using denoised.  Watching the big colored dots morph nto an image is interesting once or twice, but a cleaner image all along is much more useful.

godly-devotion:
Thanks for updating the PR. Performance is much more acceptable now.
I agree with @jrittvo we should simplify to just having denoised setting and just keep "Show Generating Image Preview".

godly-devotion:
Ah just found a bug where if you stop generation, the placeholder gallery item does not disappear.

jrittvo:
I also saw instances where at the completion of a single generation, the gallery displayed 2 images, one just being a visual duplicate.  But only one could be deleted.  The other was not even recognized by a right or left click.  I didn't check if it was backed by an actual file on dick, but I bet not.  Restarting the app removed the "phantom" one.  It was mostly happening on the first run of a session, not on subsequent runs.  I should have noted this here before.  I'm fussy about UI minutia, but bugs don't bother me.  Not helpful behavior for a tester.

hoseins77:
@godly-devotion and @jrittvo thank you for letting me know of these bugs, i will fix them soon.

jrittvo:
I am still getting the duplicate "phantom" image using the PR update from earlier today.  The 2nd image looks like it is the last update before the final image is displayed and saved.  Like a buffer or working cache that is not being removed from the image gallery when the generation cycle is completed and the final image is added to the gallery.  And lots of crashes in the Settings dialog.

The preview itself is faster now, and I am getting more image updates.  It looks great, so I hope the bugs can be fixed.  It really seems worth the trouble to do so.  At the speed it runs now, and with the clean images, I'd personally leave it on all the time.  And even if the bugs can't be resolved, I going to keep it in my own build.  Mochi takes about 2 seconds to restart on my machine.  I really don't mind the crashes. It goes right back where it was before the crash each time.

hoseins77:
The crashes i belive it is some issue of the stable-defusion  package. Adter i ypdated the package to latest version crash happend and i also tried removing my code completely to find out if uts my codes issue but it didnt resolve the problem @jrittvo

jrittvo:
Thanks for the lead. I will try to see what was added there in the past few days, and try to remove the changes one file at a time if I can, to find the one where the bug first appears.

jrittvo:
We went from using a bit past version 0.4.0 to using a bit past version 1.0.0 when the NaturalLanguage conflicts were removed for builds with Xcode 14.3.1.  That was about 5 days ago.  A lot of stuff changed from 0.4.0 to 1.0.0.  It will be very hard to figure out where the bug is with so many changes.  

It is also possible that the bug is only from the code that made the un-noised previews available 2 days ago.  I will try builds without your Mochi PR but with 1.0.0 ml-stable-diffusion, and then add in the un-noised preview changes, and hopefully see which set of changes introduces the Settings crashes.  

Assuming it is the 1.0.0 update, I will then see if the crashes are present with Xcode 15.  1.0.0 was really meant for Xcode 15.  We might be stuck with features that will be buggy until we move to Xcode 15, or we may have to revert the features until then.  It'll take me a day or two to see what this seems to be about.

jrittvo:
The move from 0.4.0 to 1.0.0 looks like it was changes to 3 existing Swift files.  There is also a new file for MultiLanguage which I don't thing Mochi ever uses, and they made changes to the Swift CLI file which is probably also irrelevant.  Of the 3 changed files that we do use, 2 are basic pipeline and the 3rd is for the text encoder.  So really that means 2 files that could be introducing the bugs.  It may not be as hard to figure this out as I first feared.  

I have a feeling that it is related to model loading/initializing or whatever it is that happens when you hit the Apply button in Settings that doesn't really need to happen.  So we might be able to skirt the bug(s) in most instances with our code cleaned up a bit.  But I am just speculating here.

hoseins77:
@jrittvo the crash happenda exactly on model init in our code and its not just on settings apply. It also happends when you change method.

jrittvo:
I have seen it also now on changing models.  So we are narrowing it down to model init.  Next step will need to be trying to find the change that introduced the bug.  I have never done it before I think you can use Xcode to move through the commit history and build and run at each point in time to find the commit that first introduced a behavior.  I will try to start on that later today.

jrittvo:
https://github.com/godly-devotion/MochiDiffusion/issues/282

I don't know where to begin reading a crash report but a few of you guys do.

jrittvo:
Info on the commit that introduced the model change crash.

https://github.com/godly-devotion/MochiDiffusion/issues/282#issuecomment-1641371128

jrittvo:
After the last commit to this PR, the Swift code error checker reports:

    Immutable value 'progress' was never used; consider removing it.

These are the diffs suggested by that comment:

`56-    if let currentImage = store.currentGeneratingImage, case let .running (progress) = generator.state`
`56+    if let currentImage = store.currentGeneratingImage, case .running = generator.state`

`92-    if let currentImage = store.currentGeneratingImage, case let .running (progress) = generator.state`
`92+    if let currentImage = store.currentGeneratingImage, case .running = generator.state`

The build is clean afterwards, and the app appears to function the same as it did before the changes.

------

There is a small bug, however, both before and after the above code clean-up.  

When the preview feature is turned off in Settings, the subsequent generation cycle displays a non-updating "preview" of the prior completed image, along with the progress counter.  Once this first new generation completes, the "preview" is no longer present in further generations, which is the correct behavior.  Enabling the preview feature again in Settings correctly returns it to the Gallery.  This on/off/on behavior is repeatable.

So there is just the one-time partial residual of the preview process in the UI that needs to be prevented when the feature is first turned off and the next run takes place.

hoseins77:
@jrittvo Thanks for testing. i think i fixed above problems in the latest commits

godly-devotion:
> Thanks for updating the PR. Performance is much more acceptable now. I agree with @jrittvo we should simplify to just having denoised setting and just keep "Show Generating Image Preview". There are some improvements to be made to the new setting but I can fix that.

I still think we should simplify to one setting with "Show Generating Image" displaying denoised previews.

jrittvo:
FWIW, I merged the latest commits from this PR into the new 4.1.2 version, along with a right-click update of the ml-stable-diffusion/main package in the dependencies list..  I did the merge manually, with my diff/merge app, because these are based on 4.1, I think.  Not sure if that was really a necessary step, but it was easy enough to do and it built with no conflicts, warnings or errors.  And to get to the point, it seems to run very well, with no crashes and the "phantom" and residual preview issues resolved.

hoseins77:
@godly-devotion I simplified the settings to just use denoised images. and uses denoised Intermediates only when preview is on i can change it to be always on but i thought it may affect the performance

jrittvo:
This PR continues to work well with Mochi 4.1.3.

godly-devotion:
Looks good, thanks!

Issue #266, opened 2023-06-20
Min value for img2img strength seems too high in latest version?


jmilr:
### Running Latest Version

- [X] I am running the latest version

### Processor

M1 (or later)

### Memory

8GB

### What happened?

I get results very different to the original image even when set to minimum (e.g. for a final pass after manual tweaking in Photoshop etc.) - is there some way to get more granular control?

jrittvo:
Can you install Mochi 3.2  for a little while to see if it gives you better control?  The range of image2image settings was changed in Mochi 4.0 because most people found that any setting of the slider from the middle to the left did not have any effect at all in moving the generated image away from the starting image.  If that change was a mistake, it can be reverted, but your experience contradicts our testing.  Setting the slider all the way to the left should pretty much ignore the text prompt and just duplicate the image, even in Mochi 4.0.

jrittvo:
I just did a test and you may be right.  It needs a little bit lower value at the low end.  If you can please try with Mochi 3.2, though, and see where the slider starts to have an effect for you, you can help us calibrate to a better level at the low end.

jrittvo:
This is what I get with Mochi 3.2.  The top left is the original image.  The top middle is the slider at 0 (all the way to the left).  Top right is at about 1.  Middle row goes 2, 3, 4.  Bottom row is 5, 6. 7.  

The minimum is presently at 6.7 if I remember right.  It should probably be moved down to 4.0.  Would that be better for you?

![Screencap](https://github.com/godly-devotion/MochiDiffusion/assets/114269461/bd63cfbb-7b1f-442f-abf2-8ae36800f82f)

jmilr:
So I'm not crazy!  - the minimum went from 0 to 6.7? Here's what I get in 4.0, without any prompt, just img2img set to lowest, and with a basic prompt
<img width="1046" alt="Screenshot 2023-06-20 at 2 07 07 pm" src="https://github.com/godly-devotion/MochiDiffusion/assets/15384770/cbf9b874-e656-4ba4-9d00-2fc47c72a081">

jrittvo:
Yep.  That matches what I see in the pics I posted above from 3.2.  The change just begins with the slider at around 4.  Moving the minimum from 0 to 6+ was too far.  A minimum of 4 would be better.  I actually did a fair amount of testing to try to pin down the minimum value that made any change, but I blew it.  Sorry!

jrittvo:
[...]

PR #279, opened 2023-07-12
Sidebar & Slider Mods

jrittvo:
This PR proposes the following changes:

1.  Adjust Image2Image STRENGTH slider minimum value down to 0.37 in response to https://github.com/godly-devotion/MochiDiffusion/issues/266

2.  Change STARTING IMAGE label to INPUT IMAGE   <---------- personal preference

3.  Reorder sidebar elements and tweak spacing.  Includes adjusting the length of the STRENGTH slider to match the length of the CONTROLNET drop down, when the sidebar is set at minimum width.  Probably it should "adjust to fill", as the drop down does, but I don't know how to code that.   (Updated it to auto-adjust, thanks to my friend ChatGPT.)   <---------- personal preference
 
No offense taken if my personal preferences don't sit well with your vision of the UI.

Screencap

### DIFF
--- a/Mochi Diffusion/Views/SidebarControls/StartingImageView.swift
+++ b/Mochi Diffusion/Views/SidebarControls/StartingImageView.swift
@@ -63,7 +63,7 @@ struct StartingImageView: View {
 
                 Slider(value: $controller.strength, in: 0.37...1, step: 0.07)
                     .controlSize(.mini)
-                    .frame(width: 110)
+                    .frame(maxWidth: .infinity)
 
                 HStack {
                     Button {
--- a/Mochi Diffusion/Views/SidebarControls/StartingImageView.swift
+++ b/Mochi Diffusion/Views/SidebarControls/StartingImageView.swift
@@ -43,7 +43,7 @@ struct StartingImageView: View {
         HStack(alignment: .top) {
             VStack(alignment: .leading) {
                 Text(
-                    "Starting Image",
+                    "Input Image",
                     comment: "Label for setting the starting image (commonly known as image2image)"
                 )
                 .sidebarLabelFormat()
@@ -61,9 +61,9 @@ struct StartingImageView: View {
                 )
                 .sidebarLabelFormat()
 
-                Slider(value: $controller.strength, in: 0.685...1, step: 0.035)
+                Slider(value: $controller.strength, in: 0.37...1, step: 0.07)
                     .controlSize(.mini)
-                    .frame(width: 88)
+                    .frame(width: 110)
 
                 HStack {
                     Button {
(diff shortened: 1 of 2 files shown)


jrittvo:
ChatGPT says to change `.frame(width: 110)` to `.frame(maxWidth: .infinity)`.  At least on the third try asking.  The first answer used UIKit and the second used Cocoa.  As little as I do know, I realized those were not right for Swift in macOS.  Giving it a larger piece of the existing code helped set it straight.  I will try the change out sometime tonight.

Mochi Diffusion/Views/SidebarControls/StartingImageView.swift
```diff
@@ -61,9 +61,9 @@ struct StartingImageView: View {
                 )
                 .sidebarLabelFormat()
 
-                Slider(value: $controller.strength, in: 0.685...1, step: 0.035)
+                Slider(value: $controller.strength, in: 0.37...1, step: 0.07)
                     .controlSize(.mini)
-                    .frame(width: 88)
+                    .frame(maxWidth: .infinity)
```
godly-devotion, line 66: Good catch on the UI here. Removing the `.frame` modifier gives the same results.

godly-devotion:
Thanks for your PR. For now I think the current control order & wording is appropriate.
However I remember on Discord that the starting image strength value range was too high, I think that is a good change to have.

jrittvo:
You're welcome to use any parts of the PR you choose.  You can just use the StartingImageView.swift file, and only the slider change(s) there.  I know rewording opens up the need for a new translations round, which makes no sense at all for 1 word.

Mochi Diffusion/Views/SidebarControls/StartingImageView.swift
jrittvo: So the default behavior is to auto adjust?  Good to know!

godly-devotion:
I can make the changes if its easier for you. If you want, you could try giving it a shot in this PR as a git exercise ;)

jrittvo:
Please make the changes.  It takes me at least 2 tries in git to get things submitted right.  I think I know how to use branches now, which will make it easier in the future, but this one is still all over the place in my repos.  I'm actually looking forward to rebasing my repo to before I submitted the PR and then doing a git pull to clean out the mess I made.

I am trying to understand the code in the Image Preview PR right now.  That will keep me busy for a while.  ml-stable-diffusion added a boolean `public var` for `useDenoisedIntermediates`.  So I think the only change to the PR for Mochi would be to set that variable to true.  It defaults to false, which I what I saw it doing last night.  But first I need to learn how to set a `public var`.  Baby steps.

godly-devotion:
Implemented via 2f8134d

Issue #282, opened 2023-07-18
App crashes when clicking apply in the settings menu


eetgeenappels:
### Running Latest Version

- [X] I am running the latest version

### Processor

M1 (or later)

### Memory

8GB

### What happened?

Go to the setting of the app and click apply

### Crash Log

```shell
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Mochi Diffusion [1909]
Path:                  /Applications/Mochi Diffusion.app/Contents/MacOS/Mochi Diffusion
Identifier:            com.joshua-park.Mochi-Diffusion
Version:               4.1 (4.1)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               502

Date/Time:             2023-07-18 17:11:46.8012 -0300
OS Version:            macOS 14.0 (23A5286g)
Report Version:        12
Anonymous UUID:        C940E3B4-B203-A1C2-A7EC-E463ACD680AA

Sleep/Wake UUID:       1B81A0E7-B793-4A84-8944-8D9D58CCCE08

Time Awake Since Boot: 2700 seconds
Time Since Wake:       2763 seconds

System Integrity Protection: enabled

Crashed Thread:        6

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000d885400ecd70 -> 0x00005885400ecd70 (possible pointer authentication failure)
Exception Codes:       0x0000000000000001, 0x0000d885400ecd70

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [1909]

VM Region Info: 0x5885400ecd70 is not in any region.  Bytes after previous region: 96848292269425  Bytes before following region: 8223787659920
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      commpage (reserved)        1000000000-7000000000   [384.0G] ---/--- SM=NUL  ...(unallocated)
--->  GAP OF 0x5f9000000000 BYTES
      MALLOC_NANO              600000000000-600008000000 [128.0M] rw-/rwx SM=PRV  

Thread 0::  Dispatch queue: com.apple.main-thread
0   AttributeGraph                	       0x1b2be46ec void AG::Graph::add_output_edge<AG::Node>(AG::data::ptr<AG::Node>, AG::AttributeID) + 68
[...]

PR #283, opened 2023-07-19
fix: set loadModel function as main actor to prevent crash

hoseins77:
In this PR i fixed model loading crash bug. happend on changing model and applying settings.
fixes: #282

### DIFF
--- a/Mochi Diffusion/Support/ImageGenerator.swift
+++ b/Mochi Diffusion/Support/ImageGenerator.swift
@@ -142,7 +142,6 @@ class ImageGenerator: ObservableObject {
         return models
     }
 
-    @MainActor
     func load(model: SDModel, controlNet: [String] = [], computeUnit: MLComputeUnits, reduceMemory: Bool) async throws {
         let fm = FileManager.default
         if !fm.fileExists(atPath: model.url.path) {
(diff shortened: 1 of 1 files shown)


jrittvo:
I then added the add Preview PR #270, with the necessary diff edits to the ImageGenerator.swift file that is changed in both PRs, and everything appears to be working smoothly in another 15 minutes of testing.

There is still the "phantom" duplicate image residual in the Gallery after each new generation completes when Image Preview is enabled.  (A bug I would typically ignore...)

godly-devotion:
I'm a bit cautious of the solution of loading the function in the main thread. I feel like its masking the real issue by forcing it to run the code sequentially. Not to mention this is now blocking the main thread.

hoseins77:
@jrittvo i have fixed preview not desapering problem i will push it soon

hoseins77:
@godly-devotion please check out new changes

jrittvo:
The alternate fix also works consistently in the build I did.

vzsg:
Two sidenotes:
- you should revert the development team change
- there's a typo: `pipline`

hoseins77:
@vzsg Thank you for noticing, i fixed them.

godly-devotion:
Thanks for the changes. I've implemented a fix which fixes the root cause of the crash via 4fe125aa3793d940c4caa293652b39edf5201fd9.

jrittvo:
Is @godly-devotion's fix basically the same as @hoseins77's fix, but applied at a deeper level in the pipeline?  A yes or no answer here is ok for me.

hoseins77:
@jrittvo no i dont think their same. i tried to fix it by changing threads to prevent not available 'self' problem, and @godly-devotion i think fixed it by removing extra function call which caused the crash.

PR #127, opened 2023-02-01
Sort models in the dropdown

vzsg:
Fixes #126.

PR #284, opened 2023-07-20
Sort ControlNet List

jrittvo:
Display list of ControlNet models alphabetically in drop-down menu


godly-devotion:
Good start! For consistency with the model selector, we could do something like this https://github.com/godly-devotion/MochiDiffusion/pull/127/files

jrittvo:
Gonna take me a bit to grasp what you are getting at, but this is the sort of thing I should be able to figure out with a little help from ChatGPT.  I will dig into it over the weekend.  Thanks for the next step suggestion!

jrittvo:
Updated ControlNet List sort method to match Model List sort method.

godly-devotion:
Looks good, just made some minor changes for completeness. Thanks!

PR #285, opened 2023-07-20
Add persistence layer

jeffreydthompson:
Discord discussion: https://discord.com/channels/1068185566782423092/1068187960350085140/1130435364855754854

Architectural changes diagram: https://www.figma.com/file/TJ5nYu1CnCaM157tSCttEZ/MochiDiffusion-proposed-Persistence-layer-change?type=whiteboard&node-id=0-1


jrittvo:
Maybe someone can help a beginner here.  Is this correct, at a very, very simple level:  The Repo sets up a central definition of where the app will look for and find models, and the Factory makes a particular model or set of models in the Repo (selected by the user in this case) available to be used by the rest of the app?  I'll take a yes or no answer here and do some reading from there if "no".

jeffreydthompson:
@jrittvo Sure, it's "a" pattern for software development.  If you follow the discord chat as well there's plenty of counterargument for this too, especially as applied to this project.  This PR may not go through, but I guess it could still be useful for learning for some.  That synopsis is correct.  It's creating a source, but the source becomes "generic" in the sense that it's just some object that conforms to some protocol.  Much like USB is a protocol, 4 wires (Pre-C): 5v+ GND D+ D- and as long as some electronic device conforms to that - it can be plugged into the computer and the computer doesn't care what it is.

The benefits of making it generic (in general - see the discord chat for the counterarguments relative to this project) is that where the source of data is pulling from is completely independent of the implementation.  It can pull from disk, Core Data, the internet, the implementation doesn't have to know.  

My work deals a lot with bluetooth - so it's useful to use a similar pattern to have one implementation that will talk to real target devices through the bluetooth radio, another that will talk to imitation test Arduino devices through the bluetooth radio, and another one that will simply be fake software devices that are sending advertising data on a timer.  

A factory would choose at runtime, usually based on some set of parameters (like using simulator - no Bluetooth radio), which concrete implementation to return.  The code that uses the implementation has no idea what was handed to it and doesn't need to know.

further reading topics:
repository/factory pattern, generics, polymorphism

jrittvo:
Thank you!  Very clearly presented.  As I was reading, the word "portability" came to mind.  I can see some advantages in clarity of the code when you use task-based modules like this.  And for complex projects, different programmers or teams can work very independently on different parts, as long as the interface syntax between modules is set.  Now I need to go back and see if the counterargument in Discord is something I can understand.

godly-devotion:
Looks like the build is failing, could you please take a look?

jeffreydthompson:
Yes, hope to revisit this weekend.  Wanted to make some changes as
discussed in the discord chat as well.

jeffreydthompson:
too many conflicts. bad merge.  will do a clean one and retry.

PR #286, opened 2023-07-20
Fix Norwegian localization not added to the project

espenbye:


godly-devotion:
Thanks for the heads up. Fixed by b3bb922.

PR #294, opened 2023-07-28
4.1 + @MainActor

jrittvo:
This is the branch that I built from, that seems to work for everything.

1.  It starts at 4.1
2.  It adds the @MainActor PR (1 line of code)
3.  It adds Image Preview, from a stash I made before your final merge.  It is slightly different from the final merge, but the differences are just cosmetic, I think.
4.  It adds the CN model list sort PR, from before the change you made when you merged it.  Also not an important difference.
5.  It reverts ml-stable-diffusion to the commit that added vzsg's InPaint PR merge.
6.  It does not include anything else that was added in Mochi on the way to 4.1.3.  There are probably a few other things that should be added, but I'm out of juice for tonight.

So this is basically meant as a guide.


jrittvo:
I think the conflicts in the set of commits above all need to have the new commit versions selected.  It is pretty clear what is what in each case.  

Moving forward, there were 4 other commits after this point (besides appcast/I10n stuff) that went into 4.1.1, 4.1.2 and 4.1.3.  I've saved each as a patch, and I will try to apply them one at a time (if I can make sense of any conflicts that need to be resolved) to see which are safe and which introduce any ControlNet bugs.  My project on and off for today.

jrittvo:
I can't make heads or tails of trying to apply and test the 4 patches that track the commits between 4.1 and 4.1.3.  It is way beyond my understanding of the code.  I could probably do it if I go back and skip merging the Image Preview stuff.  That is what is confusing me.  

It's your call.  If you want to take over from the merges in this PR that seem to let Mochi work properly and then add in some of the later merges, that's ok with me.  If you'd prefer that I try testing one merge at a time, without Image Preview in the mix, I'm pretty sure I can do that and I'd have time for that over the weekend.

jrittvo:
I am working my way through testing the merges discussed above,  I'll have a list of what worked and what didn't in a few hours.  It may not spell out an exact path forward, but I should save some time in eliminating any bad apples.

jrittvo:
Replacing this with an updated draft PR with better info.

PR #296, opened 2023-07-31
Use SDXL pipeline for XL models

ZachNagengast:
This PR just detects if the current model is an SDXL model using the presence of the new inputs `time_ids` and `text_embeds`, and uses the new pipeline `StableDiffusionXLPipeline` with the supported params. There still needs to be safety check and controlnet support, but that should happen upstream.

Just put one of these models in your model folder, and give it a try.
Normal: https://huggingface.co/apple/coreml-stable-diffusion-xl-base/blob/main/coreml-stable-diffusion-xl-base_original_compiled.zip
Palletized (Requires macOS 14): https://huggingface.co/apple/coreml-stable-diffusion-mixed-bit-palettization/blob/main/coreml-stable-diffusion-mixed-bit-palettization_original_compiled.zip

### DIFF
--- a/Mochi Diffusion.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Mochi Diffusion.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -12,10 +12,10 @@
     {
       "identity" : "ml-stable-diffusion",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/ml-stable-diffusion",
+      "location" : "https://github.com/apple/ml-stable-diffusion.git",
       "state" : {
         "branch" : "main",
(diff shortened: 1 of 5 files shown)


jrittvo:
Mochi uses a hash for the ml-stable diffusion package dependency from a few weeks ago.  I assume that needs to be changed to the hash from your SDXL merges to make this PR work, and if so, you need to add an update to the latest hash to this PR.

ZachNagengast:
@jrittvo good catch, was having issues with the cache so I left it out, but was able to get it to use the latest now

jrittvo:
Xcode 14.3 build attempt throws the errors in the attached pic.  This is probably going to need a fix similar to:

https://github.com/apple/ml-stable-diffusion/pull/208

Screencap

jrittvo:
I guess same as the failed check:

Screencap

godly-devotion:
Thanks!

Issue #290, opened 2023-07-25
Build Error: Type 'NSImage' does not conform to protocol 'Transferable'


simd-float4x4:
### Running Latest Version

- [X] I am running the latest version

### Processor

M1 (or later)

### Memory

8GB

### What happened?

Build Error **Type 'NSImage' does not conform to protocol 'Transferable'** occurs at line:54 of Extensions.swift 
Does anyone have an idea how to rewrite the code or solve the problem?

**Extensions.swift line:54**
--------------------------------
extension NSImage: Transferable { 　_**//Error: Type 'NSImage' does not conform to protocol 'Transferable'**_
    private static var urlCache = [Int: URL]()

    public static var transferRepresentation: some TransferRepresentation {
        // swiftlint:disable:next trailing_closure
        ProxyRepresentation<NSImage, URL>(exporting: { image in
            let nsImage: NSImage = image
            return try nsImage.temporaryFileURL()
        })
    }
-------------------------------

simd-float4x4:
XCode version: Version 15.0 beta 2 (15A5161b)

jrittvo:
Meaning just edit the first linet to this?

 extension NSImage: {

ZachNagengast:
Yep exactly

simd-float4x4:
Thank you for answer!
I'll try it later, and if there were some problem, I'll give you a question again.

PR #305, opened 2023-08-17
Fix transferable protocol on macOS 14

ZachNagengast:
Apple Feedback recommended against extending AppKit to conform to the transferable protocol and suggested using a wrapper. Below is their suggestion, which differs slightly from this PR in order to work well with the codebase.

This addresses issue #290 

If possible, test on both macOS 13 and macOS 14 beta.

```swift
@available(macOS, introduced: 13.0, deprecated: 14.0)
public struct TransferableImage {
    public let image: NSImage
}

extension TransferableImage: Transferable {
    public static var transferRepresentation: some TransferRepresentation {
        if #available(macOS 14.0, *) {
            return NSImage.transferRepresentation
        } else {
            return ProxyRepresentation<TransferableImage, URL>(exporting: { (transferableImage: TransferableImage) in
                return try transferableImage.image.temporaryFileURL()
            })
        }
    }
}

extension NSImage {
    fileprivate static var urlCache = [Int: URL]()
    
    fileprivate func temporaryFileURL() throws -> URL {
        let imageHash = self.getImageHash()
        if let cachedURL = Self.urlCache[imageHash] {
            return cachedURL
        }
        let name = String(imageHash)
        let url = FileManager.default.temporaryDirectory.appendingPathComponent(name, conformingTo: .png)
        let fileWrapper = FileWrapper(regularFileWithContents: self.toPngData())
        try fileWrapper.write(to: url, originalContentsURL: nil)
        Self.urlCache[imageHash] = url
        return url
    }
    
    private static func cleanupTempFiles() {
        for url in self.urlCache {
            try? FileManager.default.removeItem(at: url.value)
        }
    }
}
```

### DIFF
--- a/Mochi Diffusion/Support/Extensions.swift
+++ b/Mochi Diffusion/Support/Extensions.swift
@@ -61,6 +61,7 @@ extension TransferableImage: Transferable {
         if #available(macOS 14.0, *) {
             return NSImage.transferRepresentation
         } else {
+            // swiftlint:disable:next trailing_closure
             return ProxyRepresentation<TransferableImage, URL>(exporting: { (transferableImage: TransferableImage) in
                 return try transferableImage.image.temporaryFileURL()
             })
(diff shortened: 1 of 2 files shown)


godly-devotion:
Nice find! In regards to the `trailing_closure` feel free to remove line 156-157 from the `.swiftlint.yml` file so it just becomes an error. I don't have strong feelings towards it.

PR #318, opened 2023-10-02
Restore tiny circles as the gallery background on macOS 14

vzsg:
The following syntax no longer works on macOS 14 (nor iOS 17). The circle symbol lost its intrinsic size and instead of being repeated, a single oval fills the gallery.

```
Image(systemName: "circle.fill")
    .resizable(resizingMode: .tile)
```

I tried to look for a solution, but none worked.
So I imitated it the old background with image asset instead.


godly-devotion:
image

PR #320, opened 2023-10-05
Allow selecting models with chunked Unets

### LINKED ISSUE #319: coreml-stable-diffusion-xl-base-ios
### Processor

M1 Pro (or later)

### Memory

16GB

### What happened?

I am not sure is it bug or not, but that model is not visible in the model list
https://huggingface.co/apple/coreml-stable-diffusion-xl-base-ios

Is it possible to use "ios" models with mochi?

### Crash Log

_No response_

vzsg:
Fixes #319.

### DIFF
--- a/Mochi Diffusion/Model/SDModel.swift
+++ b/Mochi Diffusion/Model/SDModel.swift
@@ -36,15 +36,9 @@ struct SDModel: Identifiable, Hashable {
 }
 
 private func identifyAttentionType(_ url: URL) -> SDModelAttentionType? {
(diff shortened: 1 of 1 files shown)


jrittvo:
All these little fixes are really important for getting things running smoothly after all the new features and other changes flooded in.  Thank you!  I'm still pretty much seeing 1 nice fix a day at each of the involved packages.  A very worthwhile and appreciated effort.  And then maybe we get a break to just be able to play with the new stuff, rather than figure out how to play with the new stuff.  Good times ahead.

Issue #312, opened 2023-09-06
Mochi image selection hotkeys conflict with system wide text editing hotkeys 


gdbing:
### Processor

M1 (or later)

### Memory

8GB

### What happened?

⌘← is a longstanding MacOS (and iPadOS) hotkey for moving the text cursor to the beginning of the current line.
Likewise, ⌘→ moves the text cursor the the end of the line. ⌘⌫ deletes the whole line starting from the cursor.

Users expect these hotkeys to work when they are editing text in a normal system text input field but Mochi hijacks and uses them to move image selection and to delete selected image, even if a text field currently has "focus".

Image selection and deletion should be mapped to new unclaimed hotkeys, or Mochi should recognize and and respect when focus is on a text field versus the image gallery.

PR #321, opened 2023-10-08
fix: text editing hot keys interrupt by gallery hot keys

hoseins77:
In this PR i made changes to disables gallery view image selection when user is editing textfields.
Fixes: #312

### DIFF
--- a/Mochi Diffusion/Main Menu/ImageCommands.swift
+++ b/Mochi Diffusion/Main Menu/ImageCommands.swift
@@ -13,10 +13,6 @@ struct ImageCommands: Commands {
     @ObservedObject var store: ImageStore
     @ObservedObject var focusController: FocusController
 
-    private var isTextFieldFocused: Bool {
-        $focusController.negativePromptFieldIsFocused.wrappedValue || $focusController.promptFieldIsFocused.wrappedValue || $focusController.seedFieldIsFocused.wrappedValue
-    }
-
     var body: some Commands {
         CommandMenu("Image") {
             Section {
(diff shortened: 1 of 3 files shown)


Mochi Diffusion/Main Menu/ImageCommands.swift
```diff
@@ -11,6 +11,11 @@ struct ImageCommands: Commands {
     @ObservedObject var controller: ImageController
     @ObservedObject var generator: ImageGenerator
     @ObservedObject var store: ImageStore
+    @ObservedObject var focusController: FocusController
+
+    private var isTextFieldFocused: Bool {
```
godly-devotion, line 16: We should move this into the `FocusController` itself. This can be potentially useful in other places.

hoseins77: @godly-devotion do you mean isTextFieldFocused variable?

Mochi Diffusion/Main Menu/ImageCommands.swift
```diff
@@ -42,7 +47,7 @@ struct ImageCommands: Commands {
                     )
                 }
                 .keyboardShortcut(.rightArrow, modifiers: .command)
```
godly-devotion, line 49: We could probably even go further and just allow the arrow key itself without the need for the modifier key.

Example:
`.keyboardShortcut(.rightArrow, modifiers: [])`

Mochi Diffusion/Main Menu/ImageCommands.swift
```diff
@@ -53,7 +58,7 @@ struct ImageCommands: Commands {
                     )
                 }
                 .keyboardShortcut(.leftArrow, modifiers: .command)
```
godly-devotion, line 60: We could probably even go further and just allow the arrow key itself without the need for the modifier key.

Example:
`.keyboardShortcut(.leftArrow, modifiers: [])`

Mochi Diffusion/Main Menu/ImageCommands.swift
godly-devotion: Correct.

hoseins77: i will change it now

godly-devotion:
Looks good, thanks!

PR #322, opened 2023-10-10
Settings: allow selecting directories with an Open panel

vzsg:
This PR repurposes the magnifying glass buttons: instead of revealing the current directory in Finder, an Open panel is presented to pick/create directories instead. I believe this is a more useful and less confusing functionality for them.

image


godly-devotion:
Good idea!

PR #333, opened 2023-11-07
Rearrange image-to-image sidebar section

gdbing:
- lower strength slider minimum value to 0
- move stength slider below starting image selection/preview
- align info (i) button with "Strength" heading

before:
Screenshot 2023-11-07 at 3 19 50 PM

after:
Screenshot 2023-11-07 at 3 19 05 PM

The purpose of this change is to allow very low strength img2img generations. The current min value is 0.37 in part because it is too finicky and difficult to pick from the full range of 0.0 to 1.0 with such a small slider. Moving the slider below the image picker allows it to be the full sidebar width.


godly-devotion:
Looks good, thanks!

Issue #267, opened 2023-06-20
Make img2img image well an image well


jmilr:
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

I find myself dragging images into the img2img image well without thinking, and it does nothing

### Why do you think this should be added?

I just noticed it works for the control net well - is there a reason img2img can't be dragged into or clicked? Also saves having a dedicated button to add the image since you can just click a well.

![image](https://github.com/godly-devotion/MochiDiffusion/assets/15384770/6fd9f7ac-31c7-46eb-ab84-4baa5aee288b)

jrittvo:
I think this they way to go, if the empty well can display something like

Drop Starting Image
           HERE
  Or click to browse
      to an image

And the title could change to IMAGE2IMAGE.  

The CONTROLNET could be set up exactly the same, with the drop-down in place of the slider.
Simplifies everything visually, and you get a largr space for the image to display.

jmilr:
@jrittvo Yep, sounds perfect 👌 I'd only worry reducing the sidebar much more would make the prompt inputs annoying to work with, but I'm sure there's sweet-spot or a rejig of the layout.

jrittvo:
The user can resize the sidebar, above a minimum width.  The code sets the minimum width, so that could be locked in to something that makes sense for the prompt boxes, if it isn't already set that way.

jrittvo:
I realized that there is an issue with making the image wells wider.  They also get taller.  This increases the length of the sidebar column.  It will probably need scrolling, or it will need more scrolling, depending on your main window size.  I personally dislike having to scroll.  I much prefer being able to see and access everything at once, even if that means things are smaller.

PR #335, opened 2023-11-09
Allow the Starting Image to be set by drag and drop

gdbing:
fixes: #267 

This PR updates the img2img image well to support drag and drop, like the ControlNet image well already does. It doesn't remove the buttons or change the size of the image well, or any of the layout changes that were discussed in the issue page, it just allows user to drag and drop an image to load it.

Additionally
- I refactored ControlNetView to use the new ImageWellView, instead of duplicating its functionality.
- dragging an image to a well which already has an image now replaces it with the new image, previously the ControlNet image well was disabled once it was loaded with an image

### DIFF
--- a/Mochi Diffusion.xcodeproj/project.pbxproj
+++ b/Mochi Diffusion.xcodeproj/project.pbxproj
@@ -55,6 +55,7 @@
 		58FF218829FAE96200A70C7C /* ControlNetView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FF218729FAE96200A70C7C /* ControlNetView.swift */; };
 		948853A52A543CB200B4AD86 /* GalleryPreviewView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 948853A42A543CB200B4AD86 /* GalleryPreviewView.swift */; };
 		9B7392C3298DEAC2006F03D5 /* Tokenizer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B7392C2298DEAC2006F03D5 /* Tokenizer.swift */; };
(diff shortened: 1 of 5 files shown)


jrittvo:
A caveat:  There was a period of time in Mochi's development where the user making changes to one parameter (say starting image or model or scheduler) was causing other parameters to change on their own in unwanted ways.  A bug that took a month or two to find and fix.  So any change now to parameter selection (including drag and drop to an image well) will need to be very thoroughly tested before it gets released.  What was happening at that time, by way of example, was that changing the ControlNet model selection kicked out any starting image that had been set.  With no starting image info, the ControlNet model was flying blind.  I may not have it exactly right, but that is the kind of issue we need to make sure does not get introduced as a regression here.

jrittvo:
You can look at the closed PRs for ones about "MainActor" to see what was going on before that I am hoping we don't wake up again with this PR.

gdbing:
I think that Starting Image shouldn't be a problem, because the `startingImage` property is only picked up when `generate()` is called, i.e. when the Generate button is tapped. It isn't used anywhere else, so I don't think that changing it could cause unexpected side effects.

ControlNet is less straightforward to reason about because `setControlNet(CGImage)` calls `loadPipeline()` which modifies state in the `ImageGenerator`, but the new implementation is functionally identical to the old, i.e. calling `setControlNet(CGImage)` when an image is drag and dropped to it.

I've manually tested it by generating an image with consistent parameters and seed and then interrupting at different points ("ready", "loading", and "generating") by dragging and dropping new images to the starting image and control net image wells, and haven't seen any errors or variation in the resulting image. 

Is there a more comprehensive way I should test this?

jrittvo:
Sounds like you've tested it about the same way we have in the past.  I haven't had the chance to do a build with this merged, just not a good time right now, but from the little I understand of Swift, the change to MainActor gave that entity more strength against unexpected overrides.  So that fix was a real fix that should hold up.

Issue #251, opened 2023-05-04
Queue generation of multiple prompts


Naoray:
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

Being able to start multiple different prompt generations at once

### Why do you think this should be added?

Generating images from prompts takes a little while. Being able to start multiple different prompts at once would make it possible to leave the prompt generation happen while the user doesn’t have to wait in front of the pc until the first prompt generation is ready to insert a new one.

atemiz:
I think we need a queue feature in general.
Later, shortcuts that fill the queue with multiple prompt, step, guidance or seed values can be added for automation.

PR #340, opened 2023-11-22
Queued generations

gdbing:
second part to https://github.com/godly-devotion/MochiDiffusion/pull/339 addressing https://github.com/godly-devotion/MochiDiffusion/issues/251

## overview 
This PR enables user to start multiple generations without waiting for the last to finish, and replaces the progress popover in the tool bar with a list of queued generations. The queue popover design is roughly based on the Safari Downloads popover
- "Generate" button is not disabled while a generation is running
- The current generation in the JobQueueView shows a progress bar and time estimate.
- other queued generations are shown in order without progress bars
- (x) button cancels the generation
- (i) button shows parameter info for the generation
  - unused parameters are omitted, e.g. if Number of Images is 1 or if no Starting Image is selected

## localization
  - I tried to reuse strings wherever possible, so that new localization wouldn't be needed
  - In cases where new strings were needed I tried to base them on existing strings so that I could copy existing translations. 
    - new localized string "About %@" 
      - used for the time estimate under the progress bar
      - based on existing localized string "Step %lld of %lld - About %@", I just copied everything after the hyphen
    - new localized string "ControlNet Image"
      - used as accessibility label for image in the Get Info popover
      - based on existing localized strings "Image" and "ControlNet Folder"
      - Brazilian Portugese, Swedish, Spanish, and Finnish don't have translations for "ControlNet Folder". I translated it to "Imagen ControlNet" in Spanish and left it untranslated in Brazilian Portugese, Swedish, and Finnish
    - I didnt try to make translations for languages which didn't have the already localized strings I was basing the translations

## images
Screenshot 2023-11-22 at 10 37 54 AM
Screenshot 2023-11-22 at 10 38 06 AM
Screenshot 2023-11-22 at 10 40 26 AM


Mochi Diffusion/Views/JobQueueView.swift
```diff
@@ -0,0 +1,243 @@
...
+                Spacer()
+                Button("Copy Options to Sidebar") {
+                    copyOptionsToSidebar()
+                }
+            }
+        }
+    }
+}
+
+extension ImageGenerator.State: Equatable {
```
gdbing, line 225: Types observed with `onChange()` must conform to Equatable

jrittvo:
A note not directly affecting this PR:  In one of your screen caps there is a starting image in both the i2i and the CN wells. I don't think you can do both in a singe generation.  The app doesn't know to hand off the result of one pass to the other.  I think when you try to use both, it only does the i2i and it skips the CN.

jrittvo:
I merged this PR into my local build (which has a few UI tweaks) and everything seems to work well. I haven't dived into i2i or CN yet.  

There is one minor addition that I'd like to see, if it is feasible:  The seed value only lists in the queue data when it is manually set.  No entry for random seeds.  I don't know when the random seed gets assigned.  It could be well after several other queued entries are added, and I don't know if you are set up to go back and add a queue entry that would trigger a refresh of the entire queue listing.  I guess you can because the full queue listing updates when one set is removed from the queue using the "x".  I also honestly don't know how useful it would be to know the random seed for a pending generation, so this may be a silly request.

godly-devotion:
Thanks for the PR. Could you remove the local translations? Translations are done through Crowdin so any local changes will get overwritten here.

I'll also need to spend some time tweaking the UI.

gdbing:
> Could you remove the local translations? 

Done

> The seed value only lists in the queue data when it is manually set.

Random seeds aren't determined until generation begins, so unless this was changed, only the current running generation would show seed, not queued gens, and it would be possible but not trivial to pass the seed value back from ImageGenerator to show it in the UI.

It's possible to do this but I'm not convinced that it's worth it

> I don't think you can do both in a singe generation. The app doesn't know to hand off the result of one pass to the other. I think when you try to use both, it only does the i2i and it skips the CN.

It might depend on the ControlNet, but you can definitely use Shuffle ControlNet with a starting image. I believe that there has been some talk about enabling multiple ControlNets in a single generation and that the conclusion was that it would have to involve chaining them together like you describe though

Mochi Diffusion/Views/GalleryToolbarView.swift
```diff
@@ -14,55 +14,29 @@ struct GalleryToolbarView: View {
...
+                CircularProgressView(progress: stepValue)
+                    .frame(width: 16, height: 16)
+            } else if case .loading = generator.state {
                 ProgressView()
                     .progressViewStyle(.circular)
                     .controlSize(.small)
-                    .frame(width: 16)
-            }
-            .popover(isPresented: self.$isStatusPopoverShown, arrowEdge: .bottom) {
-                VStack(alignment: .leading, spacing: 4) {
```
godly-devotion, line 27: We should keep this message so that we don't give the impression to the user that generating images take a long time, but rather we are waiting for something else to complete first.

Mochi Diffusion/Views/GalleryToolbarView.swift
```diff
@@ -14,55 +14,29 @@ struct GalleryToolbarView: View {
     @State private var isStatusPopoverShown = false
 
     var body: some View {
-        if case .loading = generator.state {
```
godly-devotion, line 17: If the button doesn't do anything, we should hide it. I have button outlines visible as Accessibility option and I see a blank button with outline that doesn't do anything when clicked.

Mochi Diffusion/Views/SidebarControls/PromptView.swift
```diff
@@ -115,26 +115,17 @@ struct PromptView: View {
...
-                        )
-                    }
-                    .disabled(controller.modelName.isEmpty)
-                    .buttonStyle(.borderedProminent)
-                    .controlSize(.large)
+                Button {
+                    Task { await ImageController.shared.generate() }
+                } label: {
+                    Text(
+                        "Generate",
```
godly-devotion, line 122: After the first image started generating, it would be helpful to change the text to something like "Add to Queue".

Mochi Diffusion/Views/JobQueueView.swift
```diff
@@ -0,0 +1,243 @@
...
+                await controller.setControlNet(name: controlNetName)
+                await controller.setControlNet(image: controlNetImage)
+            } else {
+                await controller.unsetControlNet()
+            }
+        }
+    }
+
+    var body: some View {
+        VStack(alignment: .leading) {
```
godly-devotion, line 159: For this section, I would copy the style of the Inspector view for consistency (header text capitalization, spacing, etc).

godly-devotion (changes requested):

Mochi Diffusion/Views/GalleryToolbarView.swift
gdbing: Screenshot 2023-12-01 at 1 39 31 PM

Mochi Diffusion/Views/SidebarControls/PromptView.swift
gdbing: Screenshot 2023-12-01 at 1 39 33 PM

Mochi Diffusion/Views/JobQueueView.swift
gdbing: Screenshot 2023-12-01 at 1 39 46 PM

gdbing:
I addressed the comments and made one additional change, of removing the calls to "pre-load" the pipeline.

Before the refactor in https://github.com/godly-devotion/MochiDiffusion/pull/339, `loadPipeline()` was called in `ImageController` whenever the model or controlnet was changed, to load the pipeline with the new information.

The pipeline is now loaded as part of each queued job, but I left the existing calls to `loadPipeline()` with the reasoning that it could preload the pipeline so that it was already ready for the first job in a queue. My thinking has changed and I believe it is cleaner and easier to reason about if we remove those calls and only load the pipeline as part of each generation job. The meagre performance benefits aren't worth the additional complexity.

Issue #205, opened 2023-03-20
Auto resize img2img source image to the models required input size


vacekj:
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

I would love Mochi Diffusion to be able to automatically resize the selected source image when using img2img. Ideally, it would have options to crop or stretch the image, and automatically infer the correct image size from the model being used.

Keep up the great work :)

### Why do you think this should be added?

Would remove the need to manually resize source images when using img2img.

jrittvo:
If ControlNet is added to Mochi, and it supports multiple generated image sizes, this would become doubly useful.  I think I saw somewhere that Swift can resize an image from a simple command, but I don't know about cropping . . .

PR #341, opened 2023-11-27
Automatically resize input images to match model

gdbing:
fixes https://github.com/godly-devotion/MochiDiffusion/issues/205

## What
- resizes input images in img2img and ControlNet to match the model in generation pipeline
- adjusts aspect ratio of ImageWellView to show how the image will be cropped

## How
- input image dimensions are parsed from VAEEncoder.mlmodelc/metadata.json, stored as `inputSize` in SDModel
  - an early implementation I prototyped instantiated a `CoreML` `ManagedMLModel` object from the VAEEncoder.mlmodelc, to get the size, but it was too slow. Instead we directly open metadata.json and parse the size from that. This is fragile and inflexible, and will need to be updated if the model metadata ever changes, but it is more than 1000% faster
- `CGImage` extension method `scaledAndCroppedTo(size: CGSize)` is used to resize input images in generations
  - cropped to match required aspect ratio
    - so scaling won't squish or stretch image
  - crop is centred, e.g. crops equal amount off both edges
  - scaled to match required size
  - scales up or down
- applies to both img2img and ControlNet input images
- aspect ratio of ImageWellView matches the currently selected model so user can see how input image is being cropped
  - aspect ratio is square if current model doesn't support img2img (or if parsing the size from VAEEncoder.mlmodelc/metadata.json fails for any other reason)
  - image in ImageWellView is positioned in a fixed size frame so that it doesn't change size and move other controls around in the sidebar

### default square (e.g. 512x512)
Screenshot 2023-11-25 at 9 24 31 PM
Screenshot 2023-11-25 at 9 24 29 PM

### portrait (e.g. 512x768)
Screenshot 2023-11-25 at 9 24 34 PM
Screenshot 2023-11-25 at 9 24 25 PM

### Landscape (e.g. 768x512)
Screenshot 2023-11-25 at 9 24 37 PM
Screenshot 2023-11-25 at 9 24 09 PM

(NB: ImageWellView can conform to any aspect ratio, these are just the common examples)

### DIFF
--- a/Mochi Diffusion/Support/ImageController.swift
+++ b/Mochi Diffusion/Support/ImageController.swift
@@ -258,7 +258,11 @@ final class ImageController: ObservableObject {
         pipelineConfig.guidanceScale = Float(guidanceScale)
         pipelineConfig.disableSafety = !safetyChecker
         pipelineConfig.schedulerType = convertScheduler(scheduler)
-        pipelineConfig.controlNetInputs = currentControlNets.filter { $0.name != nil }.compactMap(\.image)
+        for controlNet in currentControlNets {
+            if let name = controlNet.name, let size = currentModel?.inputSize, let image = controlNet.image?.scaledAndCroppedTo(size: size) {
+                pipelineConfig.controlNetInputs.append(image)
+            }
+        }
         pipelineConfig.useDenoisedIntermediates = showGenerationPreview
 
         let genConfig = GenerationConfig(
--- a/Mochi Diffusion/Support/ImageController.swift
+++ b/Mochi Diffusion/Support/ImageController.swift
@@ -259,7 +259,7 @@ final class ImageController: ObservableObject {
         pipelineConfig.disableSafety = !safetyChecker
         pipelineConfig.schedulerType = convertScheduler(scheduler)
         for controlNet in currentControlNets {
-            if let name = controlNet.name, let size = currentModel?.inputSize, let image = controlNet.image?.scaledAndCroppedTo(size: size) {
+            if let _ = controlNet.name, let size = currentModel?.inputSize, let image = controlNet.image?.scaledAndCroppedTo(size: size) {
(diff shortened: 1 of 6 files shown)


gdbing:
from https://github.com/godly-devotion/MochiDiffusion/issues/205:
> Ideally, it would have options to crop or stretch the image, and automatically infer the correct image size from the model being used

I didn't implement the stretch option.

I can't imagine ever using it and I think that adding the option is clutter.

I'm willing to be wrong though, if folks would like to add it, it wouldn't be difficult

Mochi Diffusion/Support/ImageController.swift
```diff
@@ -249,14 +249,20 @@ final class ImageController: ObservableObject {
...
+        }
         pipelineConfig.strength = Float(strength)
         pipelineConfig.stepCount = Int(steps)
         pipelineConfig.seed = seed
         pipelineConfig.guidanceScale = Float(guidanceScale)
         pipelineConfig.disableSafety = !safetyChecker
         pipelineConfig.schedulerType = convertScheduler(scheduler)
-        pipelineConfig.controlNetInputs = currentControlNets.filter { $0.name != nil }.compactMap(\.image)
+        for controlNet in currentControlNets {
+            if let name = controlNet.name, let size = currentModel?.inputSize, let image = controlNet.image?.scaledAndCroppedTo(size: size) {
```
godly-devotion, line 262: Looks like `name = controlNet.name` is not being used and can be removed.

godly-devotion (changes requested):
Other than the minor comment, looks good!

Mochi Diffusion/Support/ImageController.swift
godly-devotion: Looks like linter is complaining about creating an unused variable. Perhaps a guard would be more appropriate (and clearer about our intentions to check for `controlNet.name` existing).

godly-devotion (approved):

Issue #337, opened 2023-11-14
Gallery should update to reflect changes to the images directory


gdbing:
### Processor

M1 Max (or later)

### Memory

64GB (or higher)

### What happened?

Files deleted (with the Finder or another app) from ~/MochiDiffusion/images/ do not disappear from the MochiDiffusion gallery until it is relaunched.

We could use `DispatchSourceFileSystemObject` to monitor changes to the folder and update the gallery to be up to date with the file system. e.g.: https://medium.com/over-engineering/monitoring-a-folder-for-changes-in-ios-dc3f8614f902

PR #345, opened 2023-12-02
Drag and Drop gallery images, and automatically stay up to date with file system

gdbing:
There are two parts to this PR

# Drag and drop images from the Gallery

- drag and drop to the ControlNet and img2img image wells
- drag and drop to the Finder to move images to a different folder
- drag and drop to Preview or Photoshop or Discord, or whatever application you wish to view or edit or use them in
  - I can drag or drop an image directly from Mochi to this GitHub PR text entry field if I want to

# Mochi watches and responds to changes in the images, models, and controlnet folders (https://github.com/godly-devotion/MochiDiffusion/issues/337)

- when images are added to or removed from the images folder, the gallery view will update accordingly
  - e.g. user can sort through and prune their generated images in the Finder or another application
- when models are added to or removed from the models folder, the model dropdown menu will update accordingly
  - removed the reload button next to the model dropdown menu, because it is no longer necessary
  - if you remove the currently selected model it selects the first model in the list
- when ControlNets are added to or removed from the controlnet folder, the ControlNet dropdown menu will update accordingly

This is a nice quality of life change, but it also helps prevent a number of edge case issues and odd behaviours which can arise when Mochi becomes "out of sync" with the file system.


godly-devotion:
Looks good! The only minor gripe is I don't know if I want the images to be removed from the gallery when they are dragged to a Finder window.

gdbing:
> I don't know if I want the images to be removed from the gallery when they are dragged to a Finder window.

I thought about that a bunch and tried out both designs and settled on drag to move (instead of drag to copy) for two reasons.

1. Because it works better with applications which use the path, such as Terminal or linking tools like Hookmark, whereas copying would require copying a temp file with an arbitrary path
2. Because `draggable` and `onDrag` automatically support system conventions like option+drag to copy, so that functionality is still available

I'm not married to my decision though, I still have the code for the other design in a branch, if we'd like to change it.

Issue #350, opened 2023-12-11
Send app notification when image generation is done


mangoes-dev:
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

I'd like there to be a Notification Centre popup when the queue is empty (i.e. all the images I queued have now been generated). Should only trigger if the app is not in foreground.

We can make this configurable and opt-in - for example only send the notification if there's been 5 or more images queued or base it on the time taken since the app was last in the foreground.

### Why do you think this should be added?

My workflow typically involves queueing up a bunch of images, then doing unrelated things on my computer until they're done. Having a notification would be awesome so I know when the images are done.

Should be do-able to do local notifications using [UNUserNotificationCenter](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter) - no servers involved.

mangoes-dev:
FWIW I'd be happy to have a go at implementing this - if you think this feature is worth having and nobody else is working on it :)

godly-devotion:
> FWIW I'd be happy to have a go at implementing this - if you think this feature is worth having and nobody else is working on it :)

Go for it 👍

PR #354, opened 2023-12-14
Send local app notifications when image generation is done

mangoes-dev:
Issue: https://github.com/godly-devotion/MochiDiffusion/issues/350

This PR adds functionality for Mochi Diffusion to send local notifications to the user when the image generation queue is done processing all the images.

- New toggle in settings 
image

- Turning on that toggle for the first time will prompt the user for access
Pasted image 20231214115442

- If the user does not select "Allow" or ignores the prompt, the app shows some helper text saying that they need to enable Notifications for Mochi Diffusion in System Settings. Clicking on the helper text takes them to the System Settings "Notifications" tab where they can grant Mochi Diffusion access to send Notifications
image

- Once the "Send Notifications" toggle is on, and Mochi Diffusion was granted access to Notifications either via the prompt or System Settings, the app now sends notifications when the job queue is empty. The notification only shows up if the app is not in foreground.
Pasted image 20231214115743


godly-devotion:
Moved Notifications under its own section in Settings but otherwise looks great!

Issue #206, opened 2023-03-23
increase steps limit


norflin321:
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

i think it would be better to limit steps to 150 (and not 50)

### Why do you think this should be added?

it will allow to produce better images, https://github.com/huggingface/swift-coreml-diffusers has 150 limit

ikolomiets:
from my experience steps in the range 50-80 keep producing hight quality variants with models that were fine-tuned with Dreambooth.

godly-devotion:
Dreambooth and Core ML produce different results based on what sampler is used. Stable Diffusion Core ML currently supports DPM Solver++ (default) and PNDM. DPM Solver++ can resolve an image with lower amount of steps than other samplers.

ten0hira:
any solution?

tjkirch:
I also often get better results beyond 50 steps.  It helps reduce wonkiness in the generated image if more specific details were requested.  Mochi Diffusion already generates images so quickly that I'd really like to wait a few more minutes for even a chance at a better result - as it is, I have no option.

PR #361, opened 2023-12-18
Sliders allow keyboard input by editing text

gdbing:
Turn the numbers on sliders into TextFields so that user can use keyboard to directly input values, making it easier to e.g. enter a precise specific value on a slider which spans from 1-100 and is only 200px wide

I had concerns about the discoverability of this feature and modified the design I originally previewed in the Discord
- the cursor changes to an [I-beam](https://support.apple.com/en-ca/guide/mac-help/mh35695/mac) when it passes over the TextField, to indicate a text-entry point
- users select the TextField by clicking once, instead of double clicking, to make it easier to stumble upon

This change addresses https://github.com/godly-devotion/MochiDiffusion/issues/206. The "Steps" and "Number of Images" sliders don't have strict upper bounds, e.g. the Steps slider only goes from 2-50, but values higher than 50 are accepted as valid input.

# UX details

- The slider updates, live, while user types into TextField

https://github.com/godly-devotion/MochiDiffusion/assets/8458547/fe97abcf-4f89-4114-a75d-828bf0915e3e

- It ignores invalid input, and uses the last valid value

https://github.com/godly-devotion/MochiDiffusion/assets/8458547/155b09ad-7e03-4026-914d-e467b944ac69

- If input can be parsed to a real number but not to a valid input (as defined by the range and step of a slider), it will be treated like the nearest valid input
  - waits until focus moves away from the TextField to "fix" the text to match the actual value

https://github.com/godly-devotion/MochiDiffusion/assets/8458547/8cac8bbc-8983-4da1-b8d4-5e9104f08740

# implementation details

## `step` Type
- `step` is Decimal instead of Double in MochiSlider init
  - Decimal is precise enough to derive fractionLength, requiring one less arg in init

## Frame Width
- add the width of an M to `textFieldSize` when a TextField is selected
  - frame should be as tight as possible, so the rest of the slider will be clickable, but need to add some width affordance to allow text input, because the TextField will redraw its content before it will resize its frame to fit the new content, causing it to stutter
  - So TextField frame needs to be wide enough to fit the current content plus one more character
  - M is wider than any digit, and than most other chars, so its width is a reasonable amount of affordance
  - There is an extensive inline comment about this

## Focus
- to manage focus, `focusedSliderField: UUID?` was added to `FocusController`
  - breaks from the existing pattern of each focusable View having its own bool, because there are multiple different simultaneous instantiations of MochiSlider

## `.onChange(of: self.value)`
  - this is a really stupid chunk of code
    - I hate it, but it works
    - there must be a much simpler and clearer way to do this though
  - Need to update the TextField when slider changes value, but shouldn't make programmatic changes to TextField while it is in use i.e. while it has focus
    - especially shouldn't recursively respond to `.onChange(of: self.text)`
  - conditional statement compares `self.text` and `self.value`
    - difference should be at least `step / 2`, to account for float imprecision
      - great enough to ignore difference between 10.0 and 10.0000001
      - but not great enough to ignore difference between 10.0 and 9.5000001 (for a step of 0.5)
    - if they aren't "significantly different" then it's probably because `self.value` was modified in `.onChange(of: self.text)`, so no need to update `text`
  - There is an extensive inline comment about this too

# Issues

## TextField overlays the Slider
TextField will intercept and block clicks from the slider. I've made the TextField frame as tight as it can reasonably be, to block as little of the slider as possible. There is space above and below the TextField to click the slider, and its possible to click the slider anywhere and drag to values covered by TextField
Screenshot 2023-12-18 at 10 58 18 AM
But a user who tries to e.g. select 0.5 on the img2img Strength slider may accidentally select the TextField instead.

## Inconsistent mouseover I-beam
When you mouseover the TextField, sometimes the cursor doesn't change into an I-beam. It still interacts normally with the TextField, selecting text if you double click or click and drag, but the cursor is a normal pointer.
There doesn't seem to be a rhyme or reason that I can discern, it seems to just (not) happen about 30% of the time.

## Input slider glitch
Sometimes, if you quickly enter two digits, the slider (blue bar) doesn't update to show the second digit. E.g. after typing 45 into its TextField, the Steps slider only appears to be 4
Screenshot 2023-12-18 at 11 35 48 AM
This seems to be some kind of race condition with updating `$value` in CompactSlider. Functionally, everything works correctly, if you generate an image it will use 45 steps, its just a visual glitch with the slider


Mochi Diffusion/Views/SidebarControls/MochiSlider.swift
```diff
@@ -0,0 +1,124 @@
...
+//
+//  Created by Graham Bing on 2023-11-08.
+//
+
+import CompactSlider
+import SwiftUI
+
+struct MochiSlider: View {
+    @EnvironmentObject private var focusCon: FocusController
+
```
godly-devotion, line 13: Unnecessary extra line.

Mochi Diffusion/Views/SidebarControls/MochiSlider.swift
```diff
@@ -0,0 +1,124 @@
...
+                self.text = value.formatted(.number.precision(.fractionLength(fractionLength)))
+            }
+        }
+        .onSubmit {
+            self.focusedSlider = nil
+        }
+        .onExitCommand {
+            self.focusedSlider = nil
+        }
+
```
godly-devotion, line 122: Unnecessary extra line.

godly-devotion (changes requested):
Although single click could help with discoverability of the function, I find that clicking lower values where the text overlaps is now impossible and requires manual number input. I think double-clicking for number input mode would be a better user experience, possibly with an outlined border like other Mac input boxes when in input mode. I'm not sure if this would cause issues where the first click sets the value then the 2nd click enters the input mode, perhaps a slight delay could mitigate that.

Overall very nice improvement!

gdbing:
> Do you think it would be possible to add an outlined border like other Mac input boxes when in input mode?

Like this?

https://github.com/godly-devotion/MochiDiffusion/assets/8458547/7268b675-c38f-4a2f-9d5b-df3625336710

I also made changes to resolve the issue I had been having with inconsistent mouseover I-beam, i.e. the cursor not always transforming correctly when hovering over the TextFields.

And I see you edited your comment about this but I changed the TextField to require double-clicking again. It still overlaps the slider, but single-clicks pass through to set the slider value, after a brief delay. I myself have gone back and forth on which is better. Both have issues, but double-click interferes the least with the existing implementation.

Try this out and see how it feels. Can easily switch back to single-click by changing line 52 from `.gesture(TapGesture(count: 2).onEnded` to `.gesture(TapGesture(count: 1).onEnded`

godly-devotion:
Hmmm after playing around it seems single clicking might be more natural despite lower values requiring dragging the slider up then over the numbers. Because we show the I-beam cursor, the natural inclination is to just click it, like regular text inputs. I've switched to the single-click. Thanks for your PR!

PR #362, opened 2023-12-20
Adds more customization to the notifications

7408647n:
Adds more customization to the notifications


godly-devotion:
Thanks for your contribution.
So I don't want to overcomplicate the notification feature. Notifications are useful for long running tasks such as when the queue is complete. Also, although the layout of the two dropdowns work in English, other languages have different word ordering which makes it hard to properly localize this.

Issue #363, opened 2023-12-20
Support 1 step generation


amitkot:
### Running Latest Version

- [X] I am running the latest version

### What do you want Mochi Diffusion to do?

Allow generation using 1 step (currently minimum is 2)

### Why do you think this should be added?

This will allow using sdxl-turbo models.
Current scheduler options are not ideal but they suffice for my needs.

PR #364, opened 2023-12-20
Feat: support 1 step generation

amitkot:
Changes the range of allowed steps to be 1-50 (currently 2-50).

Solves #363


godly-devotion:
We used to have issues with 1 step generation because `ml-stable-diffusion` crashed, but it seems it was fixed sometime.

PR #367, opened 2023-12-22
Fix guidance ',' => '.' for proper conversion replacement.

haiodo:
Guidance slider is not handling mouse clicks, caused by #361 Since Double is not expecting ',' as part of string.

### DIFF
--- a/Mochi Diffusion/Views/SidebarControls/MochiSlider.swift
+++ b/Mochi Diffusion/Views/SidebarControls/MochiSlider.swift
@@ -30,7 +30,7 @@ struct MochiSlider: View {
     private let id = UUID()
 
     func stringToDouble(_ string: String) -> Double? {
-        if let doubleValue = Double(string) {
+        if let doubleValue = Double(string.components(separatedBy: ",").joined(separator: ".")) {
             var newValue = max(doubleValue, bounds.lowerBound)
             if strictUpperBound {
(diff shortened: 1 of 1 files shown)


godly-devotion:
What problem is this fixing?

haiodo:
@godly-devotion updated description, Guidance slider is not expect ',' during parse so it not work with clicks right now.

vzsg:
In countries where the decimal separator is not `.`, the slider will display values in the correct local format, but will not parse manual input in the same format, as `Double(string)` is not locale sensitive.

For example:

1. With region set to Hungary in System Settings:

    image

1. The comma is used as the decimal separator:

    image

1. But if I manually enter `9,5` in the slider's field:

    image

1. The value gets truncated to 9.

    image

1. But if I enter 9.5, the value is updated correctly.

---

The PR works around this by manually replacing the decimal comma with a decimal point, which isn't too bad if we consider that these two are the only widespread symbols in the world.

vzsg:
Also, dragging the CN strength and guidance scale don't work **at all** in a region with decimal commas, the number just doesn't update. This PR also fixes that.

Though on a second look, I would just use `string.replacingOccurrences(of: ",", with: ".")` instead of splitting and joining.

godly-devotion:
@vzsg I see, thanks for the explanation. I've update the code with `replacingOccurrences` method.

PR #379, opened 2024-01-12
✨ Tiny UI Tweaks

Aayush9029:
- feature: Queue count added to circular progress view on the toolbar.
- ui: Separating step count and time remaining text for better readability.

| Before | After |
|---|---|
| CleanShot 2024-01-11 at 19 03 31 | CleanShot 2024-01-11 at 19 02 54 |
| CleanShot 2024-01-11 at 20 19 12 | CleanShot 2024-01-11 at 19 30 16

### DIFF
--- a/Mochi Diffusion/Views/GalleryPreviewView.swift
+++ b/Mochi Diffusion/Views/GalleryPreviewView.swift
@@ -19,16 +19,28 @@ struct GalleryPreviewView: View {
             if case let .running(progress) = generator.state, let progress = progress, progress.stepCount > 0 {
                 let step = Int(progress.step) + 1
                 let stepValue = Double(step) / Double(progress.stepCount)
-                let stepLabel = String(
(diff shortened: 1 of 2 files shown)


Aayush9029:
We would simply hide the toolbar buttons when an image is not selected. However, the original code insists on keeping the buttons and disabling them instead, so I decided to go with this approach.

godly-devotion:
Thanks for the changes. Could you please separate each change into its own PR so that they can be independently looked at? Right now all the changes are bunched together and is hard to tell what code change was for what purpose.

Aayush9029:
@godly-devotion Do you believe this modification is sufficiently compact?

godly-devotion:
So right now you have your changes in your `main` branch. I recommend doing the following.
1. Create a temporary branch, say `ui-tweaks`, off of your `main` branch.
2. Create another branch off of `main` with one of the changes, for example: `feat: add queue count to circular progress view on toolbar`.

For each Pull Request, you can reference the changes you made in `ui-tweaks` branch (using GitHub web) to create the other branches which include the changes.

Although this may seem cumbersome, it helps future us to go back in the commit history later on to see what and why a change was made. This would be harder to do if the commit was not scoped to a change and was broad an encompassing multiple changes.

PR #391, opened 2024-01-25
refactor: ImageController to new Observable macro

godly-devotion:
### Needs to be resolved
- [ ] ImageController file is too big for the compiler to resolve


gdbing:
The Observable/environment/AppStorage changes from this PR were incorporated during the broad refactor that went into v6.0

see 4d1d716584d2fe1456a7b6d424f737163aba0cce

PR #403, opened 2024-02-17
Switch to GuernikaCore

czkoko:
- New Scheduler: LCM, DPM++ 2M Karras, DPM++ SDE Karras, Euler Ancenstral

- New Model: LCM, SDXL Turbo

- New Control: T2I-Adapter

The basic functions have been completed, and there should be many minor problems, but I'm just an photographer who like programming. 
So the code will be imperfect and unprofessional. Please help continue to improve it.

https://github.com/godly-devotion/MochiDiffusion/assets/90698189/405b5434-cf82-4ba4-84fd-52033ea9937b


gdbing:
To be clear, does this break compatibility with [existing converted coreml models](https://huggingface.co/coreml-community)?

czkoko:
> To be clear, does this break compatibility with existing converted coreml models?

There is a way to be compatible with the old model, but it will affect some new functions, so if you want to change it, it is completely incompatible.

godly-devotion:
Thanks for your work, this looks promising. Unfortunately I currently don't have enough time to review and verify that all the features are working correctly, especially since there are are number of differences between `ml-stable-diffusion`. If someone wants to check and make some changes please feel free to do so.

gdbing:
It looks like the executable bit was erroneously set on every single file and should be set back. This will have the added benefit of reducing the Files Changed from 96 to 16

`git diff-tree --no-commit-id --name-only -r fb463dccf09da7275d2df3016a21fc6aef9e4991 | tr \\n \\0 | xargs -0 chmod -x`

Mochi Diffusion/Views/JobQueueView.swift
```diff
@@ -143,19 +143,19 @@ private struct InfoPopoverView: View {
 
             controller.currentModel = config.model
 
-            if let startingImage = config.pipelineConfig.startingImage {
+            if let startingImage = config.pipelineConfig.initImage {
```
gdbing, line 146: Even if `strength` is guaranteed to be non-nil when `initImage` is non-nil (is it?) it is safer and clearer to safely unwrap them both here instead of force unwrapping strength with !
```suggestion
            if let startingImage = config.pipelineConfig.initImage, let strength = config.pipelineConfig.strength {
```

and again on line 223

Mochi Diffusion/Model/SDControlNet.swift
```diff
@@ -49,7 +49,7 @@ private func identifyControlNetSize(_ url: URL) -> CGSize? {
         return nil
     }
 
-    guard let controlnetCond = inputSchema.first(where: { ($0["name"] as? String) == "controlnet_cond" }) else {
+    guard let controlnetCond = inputSchema.first(where: { ($0["name"] as? String) == "controlnet_cond" }) ?? inputSchema.first(where: { ($0["name"] as? String) == "input" }) else {
```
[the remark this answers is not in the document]
gdbing: This all might be moot, I haven't looked deeply into it but I suspect this whole method, and much of `SDControlNet.swift` is redundant with Guernika `ControlNet.swift`

e.g. this isn't useful when you can just do something like
```
let controlNet = ControlNet(modelAt: url)
let size = controlNet.sampleSize
```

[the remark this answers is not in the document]
gdbing: I would still like to understand this change though

gdbing:
I suggested a few nitpicking code changes, but I think that the real questions about this PR are much broader.

- does this seamlessly reproduce all existing Mochi functionality?
    - I haven't begun testing it myself, I've compiled this branch but haven't converted any suitable models yet
- how to handle a compatibility breaking change like this
    - don't want users to automatically update Mochi because Sparkle prompted them to and then lose the use of their models
- are there disadvantages or reasons not to move to Guernika? 
    - beyond the obvious "it's a fork maintained by one guy"

czkoko:
@gdbing 
I haven't followed these days. I haven't seen your suggestions, but I have finished all the work. You can check the code I just submitted.

da-z:
This is impressive. Can someone guide me on converting `.safetensors` models like this one https://civitai.com/models/112902/dreamshaper-xl to coreml/guernika ?

Mochi Diffusion/Resources/en-coremldata.bin
godly-devotion, line 1: What are these files for?

czkoko: hack vae, Solve multi-resolution support

gdbing: Are these just the `VAEEncoder.mlmodelc/coremldata.bin` and `VAEDecoder.mlmodelc/coremldata.bin` files from a random model converted with "variable size" enabled?

czkoko: It is equivalent to metadata.json compiled into a binary, With flexible shape information

czkoko: Even if Variable size is turned on, VAE still has a fixed shape. The conversion tool deliberately handles it like this, because Coreml has a bug, and the flexible shape consumes a lot of memory and time, which needs the method I use for the time being.

czkoko:
@godly-devotion @gdbing 
At present, I have repeatedly tested that the new code basically runs perfectly, but there are still the following problems that need to be improved:

- Some codes are not fault-tolerant.

- Whether the model supports ControlNet and T2I-Adapter is only judged when the pipeline is loaded, and not determined when the drop-down box is displayed. GuernikaModelConverter can be compatible with both at the same time.

- StartingImageView.swift: After adding ImageWellView to the last picture with different resolutions in the form of drop, MaskEditorView may have an incorrect display problem. You need to add a drop event to clear the information of the previous picture.

- The show image preview button does not work at present. Low-quality image decoding that does not affect performance is forced to be turned on. Should the switching option of high-quality decoding be added or delete the show image preview button?

godly-devotion:
> * Whether the model supports ControlNet and T2I-Adapter is only judged when the pipeline is loaded, and not determined when the drop-down box is displayed. GuernikaModelConverter can be compatible with both at the same time.

How does Guernika app (not the converter) deal with this scenario? Do they silently fail if its set but not available?

> * The show image preview button does not work at present. Low-quality image decoding that does not affect performance is forced to be turned on. Should the switching option of high-quality decoding be added or delete the show image preview button?

Is the image preview functionality not available with GuernikaCore or the existing code just needs to be adapted for it to work?

Also do you mind resolving the remaining comments that @gdbing made as well? Again, impressive work!

Mochi Diffusion/Views/JobQueueView.swift
```diff
@@ -143,19 +143,19 @@ private struct InfoPopoverView: View {
 
             controller.currentModel = config.model
 
-            if let startingImage = config.pipelineConfig.startingImage {
+            if let startingImage = config.pipelineConfig.initImage, let strength = config.pipelineConfig.strength {
                 controller.startingImage = startingImage
-                controller.strength = Double(config.pipelineConfig.strength)
+                controller.strength = Double(config.pipelineConfig.strength!)
```
jb-apps, line 148: We already have the strength, we could just use it.
```suggestion
                controller.strength = Double(strength)
```

czkoko: fix it

czkoko:
> > * Whether the model supports ControlNet and T2I-Adapter is only judged when the pipeline is loaded, and not determined when the drop-down box is displayed. GuernikaModelConverter can be compatible with both at the same time.
> 
> How does Guernika app (not the converter) deal with this scenario? Do they silently fail if its set but not available?
> 
> > * The show image preview button does not work at present. Low-quality image decoding that does not affect performance is forced to be turned on. Should the switching option of high-quality decoding be added or delete the show image preview button?
> 
> Is the image preview functionality not available with GuernikaCore or the existing code just needs to be adapted for it to work?
> 
> Also do you mind resolving the remaining comments that @gdbing made as well? Again, impressive work!

@godly-devotion  I mentioned that some problems have been fixed.

gdbing:
Variable size is a really cool hack, and I definitely think the functionality is worth the hackiness. But it can be cleaned up and made a little less hacky. Most importantly, Mochi shouldn't destructively modify user (model) data, even if it's reversible.

Currently `VAEDecoder.mlmodelc/coremldata.bin`, `VAEEncoder.mlmodelc/coremldata.bin`, `VAEDecoder.mlmodelc/model.mil`, `VAEEncoder.mlmodelc/model.mil`, and `VAEEncoder.mlmodelc/metadata.json` are overwritten with image dimension data. the `model.mil` files are backed up, but never restored by Mochi, I think if anything the backup is vestigial dev debug code from developing the `vaeDeSDXL()` functions.

If this was the only way to implement this, I wouldn't say object, but Mochi could copy and link the model files to a new /tmp/ folder, and modify and read them from there, without making changes to the original model data.

I'm implemented this change myself and I'm just testing it out, if I think it's good code I'll maybe submit it as a PR to https://github.com/czkoko/MochiDiffusion/tree/GuernikaCore

czkoko:
> Variable size is a really cool hack, and I definitely think the functionality is worth the hackiness. But it can be cleaned up and made a little less hacky. Most importantly, Mochi shouldn't destructively modify user (model) data, even if it's reversible.
> 
> Currently `VAEDecoder.mlmodelc/coremldata.bin`, `VAEEncoder.mlmodelc/coremldata.bin`, `VAEDecoder.mlmodelc/model.mil`, `VAEEncoder.mlmodelc/model.mil`, and `VAEEncoder.mlmodelc/metadata.json` are overwritten with image dimension data. the `model.mil` files are backed up, but never restored by Mochi, I think if anything the backup is vestigial dev debug code from developing the `vaeDeSDXL()` functions.
> 
> If this was the only way to implement this, I wouldn't say object, but Mochi could copy and link the model files to a new /tmp/ folder, and modify and read them from there, without making changes to the original model data.
> 
> I'm implemented this change myself and I'm just testing it out, if I think it's good code I'll maybe submit it as a PR to https://github.com/czkoko/MochiDiffusion/tree/GuernikaCore , unless there’s a better way to offer code changes to an open PR. I could just attach a patch file.

I don't think it's necessary. If you choose to use the new function of mochi, you must use GuernikaModelConverter to reconvert the model. The new model will only apply to the new version of mochi, otherwise you can stay in V5.0.

czkoko:
> > I don't think it's necessary. If you choose to use the new function of mochi, you must use GuernikaModelConverter to reconvert the model. The new model will only apply to the new version of mochi, otherwise you can stay in V5.0.
> 
> I'm not talking about user having to reconvert models, I'm talking about `hackVAE(model:)` and `modifyInputSize(url:, height:, width:)` which modify files and overwrite model data.
> 
> In order to use Mochi, user has to acquire suitable models, either downloading pre-converted models, or converting them somehow. I consider these models "user data" because the user is responsible for sourcing them and for managing the folder they're in. They're not "app data" which I would basically describe as files which are abstracted away so that the app manages them and the user is never expected to directly interact with them.
> 
> It is bad behaviour for an application to modify user data. If Preview.app put a watermark on every image it opened, that would obviously be an egregious unacceptable modification, but it would also be unacceptable for it to reencode the image to reduce file size, because that's not what you expect an image viewing app to do.
> 
> I don't think user expects Mochi to modify the models they put in ~/MochiDiffusion/models.
> 
> If Mochi had a built-in model browser, and downloaded model data to a hidden folder like ~/Library/Application Support/com.joshua-park.Mochi-Diffusion/models, I would consider that "app data" and it would be acceptable for Mochi to modify it. But if user downloads files, puts them in a user accessible folder, and only expects Mochi to read them, then Mochi shouldn't also write to them
> 
> If my abstract philosophy of system data ownership isn't enough to convince you, there are some practical concerns too
> 
> 1. Bug mitigation
> 
> If there is a bug in `hackVAE(model:)` or `modifyInputSize(url:, height:, width:)`, Mochi could "ruin" every one of a user's models.
> 
> This isn't just a matter of rigorously testing the functions, I'm confident that you've already tested thoroughly, but in the future there may be changes to Mochi, or even to GuernikaKit, which change requirements and break compatibility with `hackVAE`ed models.
> 
> Leaving the model data unmodified means such changes can be fixed with an update to `hackVAE`, without requiring user to redownload/reconvert all their models
> 
> 2. Data portability
> 
> GuernikaKit coreml models can be used with Mochi, but they can also be used with Guernika, or the command line, or any other app which uses GuernikaKit.
> 
> Modifying a model doesn't just affect it in Mochi, but in every other app or toolchain which uses the model.

Perhaps it is a good way to inform the user that the model file will be modified at the first startup and remind the user to back up VAE. If the next version of MacOS fixes the problem, these codes can be deleted. At present, this is a temporary solution, and the author of Guernika is the same solution.

godly-devotion:
> > I don't think it's necessary. If you choose to use the new function of mochi, you must use GuernikaModelConverter to reconvert the model. The new model will only apply to the new version of mochi, otherwise you can stay in V5.0.
> 
> I'm not talking about user having to reconvert models, I'm talking about `hackVAE(model:)` and `modifyInputSize(url:, height:, width:)` which modify files and overwrite model data.
> 
> In order to use Mochi, user has to acquire suitable models, either downloading pre-converted models, or converting them somehow. I consider these models "user data" because the user is responsible for sourcing them and for managing the folder they're in. They're not "app data" which I would basically describe as files which are abstracted away so that the app manages them and the user is never expected to directly interact with them.
> 
> It is bad behaviour for an application to modify user data. If Preview.app put a watermark on every image it opened, that would obviously be an egregious unacceptable modification, but it would also be unacceptable for it to reencode the image to reduce file size, because that's not what you expect an image viewing app to do.
> 
> I don't think user expects Mochi to modify the models they put in ~/MochiDiffusion/models.
> 
> If Mochi had a built-in model browser, and downloaded model data to a hidden folder like ~/Library/Application Support/com.joshua-park.Mochi-Diffusion/models, I would consider that "app data" and it would be acceptable for Mochi to modify it. But if user downloads files, puts them in a user accessible folder, and only expects Mochi to read them, then Mochi shouldn't also write to them
> 
> If my abstract philosophy of system data ownership isn't enough to convince you, there are some practical concerns too
> 
> 1. Bug mitigation
> 
> If there is a bug in `hackVAE(model:)` or `modifyInputSize(url:, height:, width:)`, Mochi could "ruin" every one of a user's models.
> 
> This isn't just a matter of rigorously testing the functions, I'm confident that you've already tested thoroughly, but in the future there may be changes to Mochi, or even to GuernikaKit, which change requirements and break compatibility with `hackVAE`ed models.
> 
> Leaving the model data unmodified means such changes can be fixed with an update to `hackVAE`, without requiring user to redownload/reconvert all their models
> 
> 2. Data portability
> 
> GuernikaKit coreml models can be used with Mochi, but they can also be used with Guernika, or the command line, or any other app which uses GuernikaKit.
> 
> Modifying a model doesn't just affect it in Mochi, but in every other app or toolchain which uses the model.

I agree, we should avoid directly modifying the models, and if needed, do that in a temp directory or cache.

Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -152,35 +153,38 @@ struct GenerationConfig: Sendable, Identifiable {
...
-                resourcesAt: model.url,
-                controlNet: controlNet,
-                configuration: config,
-                disableSafety: true,
-                reduceMemory: reduceMemory
-            )
+        await modifyInputSize(model.url, height: ImageController.shared.height, width: ImageController.shared.width)
+        let modelresource = try GuernikaKit.load(at: model.url)
+        
+        if FileManager.default.fileExists(atPath: model.url.path() + "WuerstchenPrior.mlmodelc"){
```
gdbing, line 179: What is `WuerstchenPrior.mlmodelc` and `WuerstchenPipeline`? Can this be cleaned up?

This if-else statement isn't exhaustive, maybe add a final else statement which throws an error for unsupported models (!isXL && width > 768)

Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -152,35 +153,38 @@ struct GenerationConfig: Sendable, Identifiable {
...
+            //pipeline = modelresource as? WuerstchenPipeline
+        }else if model.isXL {
+            self.pipeline = modelresource as! StableDiffusionXLPipeline
+        } else if modelresource.sampleSize.width <= CGFloat(768) {
+            self.pipeline = modelresource as! StableDiffusionMainPipeline
```
gdbing, line 184: This if-else statement isn't exhaustive (hypothetical model which fits `!isXL && width > 768` will fall through), maybe add a final else statement which throws an error for unsupported models

godly-devotion:
@gdbing 
To help unburden @czkoko, I've temporarily added you as a collaborator so that if you want, you can commit directly to the PR.

Mochi Diffusion/Support/ImageGenerator.swift
czkoko: This is to prepare for the future Wuerstchen model (Stable Cascade). GuernikaModelConverter should support it after a while.

Mochi Diffusion/Support/Functions.swift
da-z, line 1: On line 60, choose a default scheduler when the old one is no longer available (in my case "DPM-Solver++"):

        case Metadata.scheduler:
            sdi.scheduler = Scheduler(rawValue: String(value)) ?? Scheduler.dpmSolverMultistepKarras

da-z:
Also, I notice that there are issues with model folders that contain spaces:
```
There was a problem generating images: Error Domain=NSCocoaErrorDomain Code=260 "The file “model.mil” couldn’t be opened because there is no such file."
UserInfo={NSFilePath=/Users/x/MochiDiffusion/models/DreamShaper%20XL%20v2/VAEDecoder.mlmodelc/model.mil, NSUnderlyingError=0x6000002652c0
{Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```

czkoko:
> DreamShaper%20XL%20v2

Try to delete the space.

Mochi Diffusion/Support/Functions.swift
czkoko: DPM-Solver++ and DMP++ 2M are the same

da-z: When starting the app it is trying to load some previously generated images, and that is when it fails, if they were using a solved that does not exist anymore.

czkoko: I forgot to consider this problem. For the time being, you need to transfer the files in the original picture folder.

da-z: I provided a solution above. Rationale being that the default value from SDImage should be good enough to use as a default, even if not historically accurate, instead of crashing the app.

da-z:
The only issue I found is a crash when trying to use old models. It will crash here. That could be a good place to let the user know that old models are not supported any more.

img-96E24B7F

czkoko:
> The only issue I found is a crash when trying to use old models. It will crash here, after clicking on Generate button.

It is not compatible with the old model, and recommended you use GuernikaModelConverter to reconvert to support more new functions.

da-z:
> > The only issue I found is a crash when trying to use old models. It will crash here, after clicking on Generate button.
> 
> It is not compatible with the old model, and recommended you use GuernikaModelConverter to reconvert to support more new functions.

Yeah, I know. I personally managed to create a couple of new models.

da-z:
One more thing I noticed (not sure you can reproduce it on your machine): The brush is slightly off center from the mouse pointer. Say 60px to the right and 30px up from the tip of the mouse cursor.

czkoko:
> One more thing I noticed (not sure you can reproduce it on your machine): The brush is slightly off center from the mouse pointer. Say 60px to the right and 30px up from the tip of the mouse cursor.

I didn't find the reason. The mouse is in the center of the brush on my computer.

da-z:
> > One more thing I noticed (not sure you can reproduce it on your machine): The brush is slightly off center from the mouse pointer. Say 60px to the right and 30px up from the tip of the mouse cursor.
> 
> I didn't find the reason. The mouse is in the center of the brush on my computer.

Thought so. You would have spotted it.

I clicked the 4 corners of the image, here are the results. The top right corner was spot-on, dead-center. The others not.

img-5kzyxh4J

czkoko:
> Thought so. You would have spotted it.
> 
> I clicked the 4 corners of the image, here are the results. The top right corner was spot-on, dead-center. The others not.

I also have a similar problem. Can you check it for me at StartingImageView?

czkoko:
@da-z I fixed it simply, but it doesn't seem to be perfect.

da-z:
> @da-z I fixed it simply, but it doesn't seem to be perfect.

I can confirm the brush is in the center now but the whole window is a bit too big (goes outside my monitor area and can't see the slider anymore). Can the whole preview window be scaled down a bit by some percentage? Or fix a w and h for it?

czkoko:
> I can confirm the brush is in the center now but the whole window is a bit too big (goes outside my monitor area and can't see the slider anymore). Can the whole preview window be scaled down a bit by some percentage? Or fix a w and h for it?

I didn't find a solution. It was also my first contact with Swiftui.

da-z:
While still not perfect, this seems to work better. I can live with that

img-n2Xm3Pgg

da-z:
We got it:

```swift
        MochiSlider(value: $controller.strength, bounds: 0.0...1.0, step: 0.05)
            .popover(isPresented: self.$isMaskPopoverShown, arrowEdge: .top) {
                let scalingFactor: Double = 0.8
                MaskEditorView(startingImage: controller.startingImage?.scaledAndCroppedTo(
                    size: CGSize(
                        width: (Double(controller.width) * scalingFactor).rounded(),
                        height: (Double(controller.height) * scalingFactor).rounded()
                    )
                ), maskImage: $maskImage)
            }
```

```swift
    var body: some View {
        VStack(alignment: .trailing){
            ZStack {
                if let startingImage = startingImage {
                    Image(nsImage: NSImage(cgImage: startingImage, size: NSSize(width: startingImage.width, height: startingImage.height)))
                    //.resizable()
                    //.aspectRatio(contentMode: .fit)
                }
                if let maskImage = maskImage {
                    Image(nsImage: maskImage)
                    //.resizable()
                    //.aspectRatio(contentMode: .fit)
                }
            }
```

czkoko:
@da-z I have scaled for different screen resolutions, but it seems that there are still bugs in it. After modifying it beyond the original resolution, the canvas will not be automatically updated. You must click again.

da-z:
From my point of view, this is good to be merged. Awesome work.

gdbing:
Despite my temporary collaborator status I'm able to commit changes to one file at a time with the Github web UI but don't have permission to push changes to @czkoko 's branch 

so I've created a PR to address the concerns I brought up regarding copying the models to temp folders so we don't  and pointed it to @czkoko 's branch 
https://github.com/czkoko/MochiDiffusion/pull/1
you can see the details there, let's address more any questions or discussion in that PR

czkoko:
> I'm able to make changes to this PR one file at a time with the Github web UI but despite my temporary collaborator status don't have permission to push changes from my local copy to @czkoko 's branch
> 
> so I've created a PR to address the concerns I brought up regarding copying the models to temp folders so we don't and pointed it to @czkoko 's branch [czkoko#1](https://github.com/czkoko/MochiDiffusion/pull/1) you can see the details there, let's address more any questions or discussion about this specific change in that PR

I tested your PR, but I can't merge it.
I agree that it will be more reasonable for you to move the modified VAE code to SDModel.

At present, I find that your plan has the following major problems:
- 1. Maybe the APFS file system, as you said, does not really copy files to the temporary directory, but it is limited to the same disk. If the user model is stored in a mobile SSD, the huge model file will be copied to the system temporary directory every time the model or resolution is switched.
- 2. Your scheme will cause you to reload the model once you switch the resolution, and the time it takes for the first step preview to appear is 2-3 times that of my plan.
- 3. There is a serious memory leak problem. Once the resolution changes, the memory required by mochi will almost double. The following figure shows the memory usage after switching the resolution 4 times, It has exceeded 25GB.

cchance27:
Any chance that GuernikaCore can lead to ipadapter support?

czkoko:
@godly-devotion 
At present, my work has been completed, no bugs have been found for the time being, and I have no ability to add new functions.
If possible, I hope that someone can add pre-processing functions to the input picture of ControlNet, such as using Vision API to generate human pose, MiDas model to generate depth maps, etc.

gdbing:
@czkoko Style Prompt Presets shouldn't be bundled together with GuernikaCore. It seems like an unrelated feature which should be in its own PR

Sorry for the delay, I've been busy but I'll have some additional feedback shortly. I don't think this PR should be merged yet

godly-devotion:
Agreed. We appreciate your work but there are a number of unrelated UI changes that should be reverted, such as the Gallery UI changes. Also there are still some lint errors present that needs to be fixed.

da-z:
I am very happy with Andy's work so far and I would not revert a thing (as a matter of fact, I created a fork, just in case).
Regarding the lint errors: can't those be fixed automatically? From what I see a lot of them are related to formatting.

Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -152,35 +153,46 @@ struct GenerationConfig: Sendable, Identifiable {
...
-        } else {
-            self.pipeline = try StableDiffusionPipeline(
-                resourcesAt: model.url,
-                controlNet: controlNet,
-                configuration: config,
-                disableSafety: true,
-                reduceMemory: reduceMemory
-            )
+        if model.allowsVariableSize && vaeAllowsVariableSize(model.url) == false{
+            await modifyInputSize(model.url, height: ImageController.shared.height, width: ImageController.shared.width)
```
gdbing, line 180: All of these references to ImageController are bugs. `loadPipeline` is called at the beginning of each generation job and by the time a job reaches the front of the queue, ImageController may have changed.

Plus it's just good practice to reduce coupling as much as possible. 

These values can be obtained from genConfig, but it also might make more sense to all of this cache logic (including the `hasher` logic right above) out of `loadPipeline` altogether and into `ImageController.runGenerationJobs()`  I'm imagining something like
```
if pipelineNeedsToBeReloaded {
    try await ImageGenerator.shared.loadPipeline()
}
```

Mochi Diffusion/Model/SDModel.swift
```diff
@@ -130,3 +136,281 @@ private func identifyInputSize(_ url: URL) -> CGSize? {
...
+    }
+
+    if inputSchema.first(where: { ($0["hasShapeFlexibility"] as? String) == "1" }) != nil {
+        return true
+    }else{
+        return false
+    }
+}
+
+public func vaeAllowsVariableSize(_ url: URL) -> Bool? {
```
gdbing, line 202: `vaeAllowsVariableSize` should be a private function which is used to set a `SDModel` property during init. All the places this function is called should just access that property instead.

Mochi Diffusion.xcodeproj/project.pbxproj
```diff
@@ -390,32 +398,13 @@
...
-			);
-			inputPaths = (
-			);
-			outputFileListPaths = (
-			);
-			outputPaths = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "export PATH=\"$PATH:/opt/homebrew/bin\"\nif which swiftlint > /dev/null; then\n  swiftlint\nelse\n  echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
```
gdbing, line 415: SwiftLint should be reenabled.

gdbing (changes requested):

nosferatu500:
> @godly-devotion At present, my work has been completed, no bugs have been found for the time being, and I have no ability to add new functions. If possible, I hope that someone can add pre-processing functions to the input picture of ControlNet, such as using Vision API to generate human pose, MiDas model to generate depth maps, etc.

I've converted MiDas CoreML model if needed, but I don't know anything about the MochiDiffusion codebase to implement depth estimation preprocessing, but I think you can implement it yourself, it's literally just a few lines of code to work with this model. but how to integrate it with the Mochi codebase... that's another question.

Mochi Diffusion/Model/SDControlNet.swift
```diff
@@ -8,21 +8,28 @@
...
 struct SDControlNet {
     let name: String
     let url: URL
     let size: CGSize
     let attention: SDModelAttentionType
+    let controltype: ControlType
 
     init?(url: URL) {
-        guard let size = identifyControlNetSize(url), let attention = identifyControlNetAttentionType(url) else {
+        guard let size = identifyControlNetSize(url), let attention = identifyControlNetAttentionType(url), let type = identifyControlNetType(url) else {
```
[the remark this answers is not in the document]
czkoko: ControlNet and T2IAdapter can be compatible with any shape in the conversion using variable resolution.

Mochi Diffusion/Model/SDImage.swift
```diff
@@ -21,11 +21,11 @@ struct SDImage: Identifiable, Hashable {
...
     var aspectRatio: CGFloat = 0.0
     var model = ""
-    var scheduler = Scheduler.dpmSolverMultistepScheduler
+    var scheduler = Scheduler.dpmSolverMultistepKarras
     var mlComputeUnit: MLComputeUnits?
     var seed: UInt32 = 0
-    var steps = 28
-    var guidanceScale = 11.0
+    var steps = 15
+    var guidanceScale = 5.0
```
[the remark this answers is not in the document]
czkoko: The new parameters can ensure the generation of acceptable pictures with the minimum number of steps.

Mochi Diffusion/Support/ImageController.swift
```diff
@@ -255,23 +267,52 @@ final class ImageController: ObservableObject {
...
+            pipelineConfig.size = model.inputSize
+        }
+        
+        if let size = pipelineConfig.size, startingImage != nil{
+           pipelineConfig.initImage = startingImage?.scaledAndCroppedTo(size: size)
+            pipelineConfig.inpaintMask = maskImage?.scaledAndCroppedTo(size: size)
+        }
+        if startingImage == nil && currentControlNets.isEmpty{
+            strength = 1.0
         }
```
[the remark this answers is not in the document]
czkoko: Txt2img, img2img, ControlNet, are all affected by strength. If the value is low, the number of steps will be reduced accordingly, so txt2img must be forced to 1.0

Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -152,35 +153,46 @@ struct GenerationConfig: Sendable, Identifiable {
...
+            self.pipeline = modelresource as! StableDiffusionXLPipeline
+        case is StableDiffusionXLRefinerPipeline:
+            self.pipeline = modelresource as! StableDiffusionXLRefinerPipeline
+        case is StableDiffusionPix2PixPipeline:
+            self.pipeline = modelresource as! StableDiffusionPix2PixPipeline
+        default:
+            self.pipeline = modelresource as! StableDiffusionMainPipeline
+        }
+    
+        self.pipeline?.reduceMemory = true//reduceMemory
```
gdbing, line 195:
```suggestion
        self.pipeline?.reduceMemory = reduceMemory
```

SpiraMira:
@godly-devotion - is it safe for me to assume that this branch will **not** be merged into the main branch?  Congratulations to @czkoko for this great integration effort, but I am a little worried about the level of support for GuernaKit.  It seems like it is only one developer and the changes are few and far between.  Am I wrong?

Mochi Diffusion/Support/ImageGenerator.swift
czkoko: ReduceMemory must be forced to be turned on here, otherwise img2img will not be able to use variable resolution. ReduceMemory will unload VAEEncoder to read the new input shape. Unload VAEEncoder is not public.

gdbing: Does this also apply to ControlNets and T2IAdapters?

vzsg:
> @godly-devotion - is it safe for me to assume that this branch will **not** be merged into the main branch? Congratulations to @czkoko for this great integration effort, but I am a little worried about the level of support for GuernikaKit. It seems like it is only one developer and the changes are few and far between. Am I wrong?

To be fair, Apple's ml-stable-diffusion library – that Mochi is using on main – has been even more stagnant.

Mochi Diffusion/Support/ImageGenerator.swift
czkoko: > Does this also apply to ControlNets and T2IAdapters?

Judging from the source code of Guernikakit, reduceMemory will unload ControlNets and T2IAdapters if necessary.
I compared and tested the completion time of the first step of ReduceMemory in the on and disabled state, which is almost the same.
txt2img switching resolution will not affect the speed.
img2img switching resolution requires reloading the pipeline, which takes 1-2 seconds.

SpiraMira:
> > @godly-devotion - is it safe for me to assume that this branch will **not** be merged into the main branch? Congratulations to @czkoko for this great integration effort, but I am a little worried about the level of support for GuernikaKit. It seems like it is only one developer and the changes are few and far between. Am I wrong?
> 
> To be fair, Apple's ml-stable-diffusion library – that Mochi is using on main – has been even more stagnant.

I agree.  But there have been at least 16 commits on Apple's ml-stable-diffusion library since release 1.1.0 in September (and an arguably larger community of interested parties and contributors).  GuernikaKit is chock full of good stuff (that I am inspecting for my own purposes). Breaking model compatibility is a concern but I need to deep dive into the Guernika model conversion scripts a little more before I feel comfortable...

gdbing:
I've created another PR pointed at czkoko/GuernikaCore: https://github.com/czkoko/MochiDiffusion/pull/4

It attempts to address most of the issues which I and others have raised here.

To try to keep it readable I did not include a merge of the latest upstream changes from godly-devotion/main in the PR, but it's a nontrivial merge and can be seen [here](https://github.com/gdbing/MochiDiffusion/tree/GuernikaCore%2Bfixes%2Bupstream)

gdbing:
Something odd I've found is that if I use the variable size hack to resize an SDXL model so that one dimensions is at least 1024, and the other is greater than 1024, it consistently crashes on the final processing step.

I can generate 896x1536 and 1024x1024, but if I try 1024x1152, it will crash. 

These are the only crashes I'm still seeing, and I can reproduce this going back to the earliest variable size hack builds, so it seems to be something inherent to the hack.

Does Guernika itself allow you to generate 1152x1152 SDXL images?

czkoko:
> Something odd I've found is that if I use the variable size hack to resize an SDXL model so that one dimensions is at least 1024, and the other is greater than 1024, it consistently crashes on the final processing step.
> 
> I can generate 896x1536 and 1024x1024, but if I try 1024x1152, it will crash.
> 
> These are the only crashes I'm still seeing, and I can reproduce this going back to the earliest variable size hack builds, so it seems to be something inherent to the hack.
> 
> Does Guernika itself allow you to generate 1152x1152 SDXL images?

Yes, I have found this problem for a long time. The width and height cannot exceed 1024 at the same time.
I'm sure it's caused by coremldata.bin. Guernika app does not have this problem. It modifies coremldata.bin every time it modifies the resolution. It uses coremldata.bin to compile fixed shape information. 
I tried this scheme, but failed. Using binary mode to modify the shape information caused a coreml core reading error. I don't know how the author of Guernika modified it.
Later, I found the current solution: the information of flexible shapes is compiled in coremldata.bin. The advantage is that you only need to replace the coremldata.bin once, and the disadvantage is that the width and height cannot exceed 1024 at the same time.

SpiraMira:
> > Something odd I've found is that if I use the variable size hack to resize an SDXL model so that one dimensions is at least 1024, and the other is greater than 1024, it consistently crashes on the final processing step.
> > I can generate 896x1536 and 1024x1024, but if I try 1024x1152, it will crash.
> > These are the only crashes I'm still seeing, and I can reproduce this going back to the earliest variable size hack builds, so it seems to be something inherent to the hack.
> > Does Guernika itself allow you to generate 1152x1152 SDXL images?
> 
> Yes, I have found this problem for a long time. The width and height cannot exceed 1024 at the same time. I'm sure it's caused by coremldata.bin. Guernika app does not have this problem. It modifies coremldata.bin every time it modifies the resolution. It compiled fixed shape information to coremldata.bin. I tried this scheme, but failed. Using binary mode to modify the shape information caused a coreml core reading error. I don't know how the author of Guernika modified it. Later, I found the current solution: the information of flexible shapes is compiled in coremldata.bin. The advantage is that you only need to replace the coremldata.bin once, and the disadvantage is that the width and height cannot exceed 1024 at the same time.

Hi - the hack just touches the .mil files.  Why is coremldata.bin also being backed up, renamed and copied?  I’ve searched the code here and don’t see where coremldata.bin comes into play.  Also is there a description of what coremldata.bin contains? I know that it’s a product of the coreml compiler but I can’t find any decent documentation on it (beyond it containing some metadata...) 

thanks in advance.

czkoko:
> Hi - the hack just touches the .mil files. Why is coremldata.bin also being backed up, renamed and copied? I’ve searched the code here and don’t see where coremldata.bin comes into play. Also is there a description of what coremldata.bin contains? I know that it’s a product of the coreml compiler but I can’t find any decent documentation on it (beyond it containing some metadata...)
> 
Coremldata.bin stores some information about the following picture, and CoreML Framework will read it to verify the model.

SpiraMira:
> > Hi - the hack just touches the .mil files. Why is coremldata.bin also being backed up, renamed and copied? I’ve searched the code here and don’t see where coremldata.bin comes into play. Also is there a description of what coremldata.bin contains? I know that it’s a product of the coreml compiler but I can’t find any decent documentation on it (beyond it containing some metadata...)
> 
> Coremldata.bin stores some information about the following picture, and CoreML Framework will read it to verify the model. 

thanks - a binary version of InputSchema and OutputSchema latent descriptions in metadata.json (I guess)
- how did you get that picture?
- is it being backed up just as a precaution?  (I don’t think the app modifies it in any way)

lost-illusion:
@czkoko Hi!
Im trying to test your PR, clone branch, etc, then convert model like this 
изображение

and... on my m1 max 24gpu im waiting three hours for loading model first time, and nothing. Maybe im doing something wrong ? im trying both branches, main and GuernikaCore

czkoko:
@lost-illusion 
Using cpu&gpu to reconvert the model, ANE does not support such a high resolution as 1024x1024.

lost-illusion:
> @lost-illusion Using cpu&gpu to reconvert the model, ANE does not support such a high resolution as 1024x1024.

okey, thank you, something like this ? 
изображение
will be good to have any example, for simplified testing)

czkoko:
@lost-illusion 
Recommended settings:


More instructions for use:
https://discord.com/channels/1068185566782423092/1068187870537449532/1208434036574396477

lost-illusion:
> @lost-illusion Recommended settings: 
> 
> More instructions for use: https://discord.com/channels/1068185566782423092/1068187870537449532/1208434036574396477

ty

Mochi Diffusion/Model/Scheduler.swift
```diff
@@ -5,21 +5,33 @@
...
     /// Scheduler that uses a second order DPM-Solver++ algorithm
-    case dpmSolverMultistepScheduler = "DPM-Solver++"
+    case dpmSolverMultistep = "DPM++ 2M"
+    
+    case dpmSolverMultistepKarras = "DPM++ 2M Karras"
+    
+    case dpmSolverSinglestepKarras = "DPM++ SDE Karras"
+    
+    case eulerAncenstralDiscrete = "Euler Ancenstral"
 }
```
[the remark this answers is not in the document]
alfuyao1986: These are more close to standard schedulers to hugging faces API uses. And supports more models like STXL-Turbo.

PR #405, opened 2024-03-02
Improved metadata

Dalchrome:
I have improved and added to the metadata capabilities of MochiDiffusion so metadata can be ported between apps via the saved images.

I have kept the remaining Metadata and added various others including Xmp.exif.UserComment, this is a json formatted string that matches with the A1111 api protocol so as to be compatible with CivitAI, all the main python generators, popular Mac apps and makes it easier link to other apps later too.

the main file added is MetadataHelper.swift which is my ported code as well as the existing metadata merged. This file is east to maintain and I can see plenty to add later when the transition to Guernika is complete. 

the other change is amending the SDImage.swift imagedata function, to call for the metadata


godly-devotion:
You have changes from another PR in here. Please separate and commit only your changes.

Dalchrome:
Sorry, I'm quite new to using GitHub, part of my commit needs guernikakit for the different scheduler name formatting, so I was working from czkoko fork. I can't find a way to just commit my own changes, I was hoping that the other request would complete first. Either way I will submit a new pull request soon.

PR #414, opened 2024-03-14
feat: implement progress bar and update progress indicators

alexey-detr:
Some improvements for UI around progress indicators.

Now there are no overlays over the preview image, and it's easier to stop generation, since stop button is right in the toolbar.

I don't have much Swift experience, will be glad for any hints or fix suggestions.

https://github.com/godly-devotion/MochiDiffusion/assets/334228/63b9275e-2f02-4814-af30-681ec1485ac3

### DIFF
--- a/Mochi Diffusion/Support/ImageGenerator.swift
+++ b/Mochi Diffusion/Support/ImageGenerator.swift
@@ -27,7 +27,6 @@ struct GenerationConfig: Sendable, Identifiable {
 }
 
 @Observable public final class ImageGenerator {
-
     static let shared = ImageGenerator()
 
     enum GeneratorError: Error {
@@ -88,8 +87,7 @@ struct GenerationConfig: Sendable, Identifiable {
         let items = try fm.contentsOfDirectory(
             at: finalImageDirURL,
             includingPropertiesForKeys: nil,
-            options: .skipsHiddenFiles
-        )
+            options: .skipsHiddenFiles)
         let imageURLs =
             items
             .filter { $0.isFileURL }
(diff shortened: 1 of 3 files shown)


Mochi Diffusion/Support/ImageGenerator.swift
```diff
@@ -167,7 +167,7 @@ struct GenerationConfig: Sendable, Identifiable {
         return models
     }
 
-    func loadPipeline(
+    func preparePipeline(
```
alexey-detr, line 170: I haven't found a more elegant way to handle this. Issue is that when load pipeline finished and generation of image is about to start, there's a short time while generator returns back to ready state, which makes UI flicker.

Mochi Diffusion/Views/GalleryItemView.swift
```diff
@@ -77,7 +77,6 @@ struct GalleryItemView: View {
                 Image(image, scale: 1, label: Text(verbatim: String(sdi.seed)))
                     .resizable()
                     .aspectRatio(contentMode: .fit)
-                    .padding(4)
```
alexey-detr, line 80: Not sure if it's better, but feels like this gap was unnecessary, I can revert it.

gdbing:
- How does it look in smaller windows? Not even absurdly tiny, but e.g. a 3/4 screen width window on a 13” Macbook Air at default screen resolution with both sidebars on?
- I would expect info button to show info about the current job, not a list of upcoming queued jobs. I don’t have an obvious answer for what would be a better icon though.

alexey-detr:
@gdbing Thanks for looking into this!

> How does it look in smaller windows? Not even absurdly tiny, but e.g. a 3/4 screen width window on a 13” Macbook Air at default screen resolution with both sidebars on?

Items start to hide from right to left behind `>>` button, I think it's like a native behavior of toolbar in SwiftUI. So the new progress block disappears last. I'd prefer this progress block to expand and take the whole width actually when there's some space. But that's where perhaps I need some help as I don't have much SwiftUI expertise and not sure if it's even doable.

> I would expect info button to show info about the current job, not a list of upcoming queued jobs. I don’t have an obvious answer for what would be a better icon though.

Yeah, good point, I'll see what I can do, maybe reorganizing it a little bit. The point is I wanted to provide info about current progress right in the new progress block, but the whole queue (and maybe more detailed) information would be hidden in the popup. I think this `list.bullet` icon list bullet would fit just fine for this then.

appinteractive:
Any updates?

PR #441, opened 2024-07-20
feat: define Python environment for model conversion

vlad-ivanov-name:
- [x] I have read [CONTRIBUTING.md](https://github.com/MochiDiffusion/MochiDiffusion/blob/main/CONTRIBUTING.md)

### Description:

Apple doesn't pin dependencies in their repo, so over time their code
breaks. This PR defines an environment where dependency versions are
constrained and pinned. The resulting environment should be used with
https://github.com/astral-sh/uv tool

### DIFF
--- a/conversion/download-script.sh
+++ b/conversion/download-script.sh
@@ -2,4 +2,5 @@
 
 HF_DIFFUSERS_REVISION=76c645d3a641c879384afcb43496f0b7db8cc5cb
 
-wget https://raw.githubusercontent.com/huggingface/diffusers/${HF_DIFFUSERS_REVISION}/scripts/convert_original_stable_diffusion_to_diffusers.py
+curl -L -o convert_original_stable_diffusion_to_diffusers.py \
+  https://raw.githubusercontent.com/huggingface/diffusers/${HF_DIFFUSERS_REVISION}/scripts/convert_original_stable_diffusion_to_diffusers.py


gdbing:
I'm not familiar with `uv`, but I tried setting this up and doing a conversion.

```zsh
brew install uv
cd ./conversion
uv run python --version
```
Here I hit my first issue: 

> Failed to parse lockfile; ignoring locked requirements: TOML parse error at line 723, column 35
    |
723 |     { name = "diffusers", extra = "torch" },
    |                                   ^^^^^^^
invalid type: string "torch", expected a sequence

I changed line 723 to ` { name = "diffusers", extra = ["torch"] },` and ran it again

>   × No solution found when resolving dependencies:
  ╰─▶ Because only python-coreml-stable-diffusion==1.1.0 is available and
      python-coreml-stable-diffusion==1.1.0 depends on diffusers==0.30.2, we can conclude that all versions of
      python-coreml-stable-diffusion depend on diffusers==0.30.2.
      And because your project depends on diffusers<=0.24.0 and python-coreml-stable-diffusion, we can
      conclude that your project's requirements are unsatisfiable.

It appears that `python-coreml-stable-diffusion` isn't fixed to a specific version, and the latest's `diffusers` version requirement has been updated since this PR was created.

It might have been better to fix `python-coreml-stable-diffusion` to an older version, but instead I updated `"diffusers<=0.30.2",` in pyproject.toml and ran it again. 

Now it ran successfully. Let's try converting a safetensors to mlmodelc

```zsh
mv convert_original_stable_diffusion_to_diffusers.py ./conversions
mv dreamshaper_8.safetensors ./conversions
cd ./conversions
uv run python convert_original_stable_diffusion_to_diffusers.py --checkpoint_path analogDiffusion_10Safetensors.safetensors --from_safetensors --device cpu --extract_ema --dump_path analogDiffusion_10Safetensors_diffusers
```

> Traceback (most recent call last):
  File "convert_original_stable_diffusion_to_diffusers.py", line 160, in <module>
    pipe = download_from_original_stable_diffusion_ckpt(
  File "/Users/graham.bing/Developer/MochiDiffusion/conversion/.venv/lib/python3.8/site-packages/diffusers/pipelines/stable_diffusion/convert_from_ckpt.py", line 1244, in download_from_original_stable_diffusion_ckpt
    raise ValueError(BACKENDS_MAPPING["omegaconf"][1])
ValueError:
{0} requires the omegaconf library but it was not found in your environment. You can install it with pip: `pip
install omegaconf`

So what this says to me is that bumping up the diffusers version was a mistake, but in any case, this is where I stopped.

gdbing:
Before we can merge this I need this to be resolved. I would also like a brief writeup of setup/usage instructions which could be added to https://github.com/MochiDiffusion/MochiDiffusion/wiki/How-to-convert-Stable-Diffusion-models-to-Core-ML

PR #451, opened 2024-10-08
feat: right-click to set image file Finder color tags, shown as emoji

jupdike:
When browsing generated images, right-click and tag a particular image with a star (yellow), heart (red), flame (orange), and more (7 emoji/colors total). Tags are stored natively in macOS and show up in Finder (implemented as decades-old HFS+/APFS Finder tag colors, works on iOS and iPadOS too). (The emoji are arbitrary but the color of emoji need to match Apple's predefined seven colors for this system to make sense.)

### DIFF
--- a/Mochi Diffusion/Model/ImageStore.swift
+++ b/Mochi Diffusion/Model/ImageStore.swift
@@ -116,6 +116,12 @@ enum ImagesSortType: String {
         }
     }
 
+    func updateMetadata(_ sdi: SDImage, colorNumber: Int) {
+        guard let index = index(for: sdi.id) else { return }
+        allImages[index] = sdi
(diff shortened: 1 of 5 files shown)


gdbing:
This is great, using an emoji for each colour is a fun idea. I'm just going to make a couple changes

- move Tags into a submenu, reword the menu items
- change some of the emojis
  - I went back and forth about whether to use "meaningful" emoji (❤️⭐️🍆) or just "random" emoji which match the colours, and settled on the latter

In the future it would be a nice improvement to use `kMDItemUserTags` instead of `kMDItemFSLabel`. This would allow images to have multiple different tags, as well as arbitrary string tags, like Finder currently supports, but I think this is fine for now.

PR #453, opened 2024-10-12
feat: implement advanced filter text field with filtering logic

hoseins77:
- [x] I have read [CONTRIBUTING.md](https://github.com/MochiDiffusion/MochiDiffusion/blob/main/CONTRIBUTING.md)

### Description:
Hello, this PR adds an advanced filter text field to the project, inspired by Xcode logs filtering. The new filter enhances the gallery search functionality by allowing users to refine results based on multiple criteria, such as prompt model, seed, steps count and more. It supports filter options, including "is equal" and "is not equal" conditions, as well as specifying whether the search should match exactly (equal) or simply contain the given text. These additions improve the flexibility and precision of searches within the gallery.

Screenshot 2024-10-12 at 3 53 10 AM


gdbing:
This is really great!

Arrow keys don't move the insertion point in the "Filter" TextField, but that's a preexisting issue, so I won't hold this up for it

PR #462, opened 2024-11-16
feat: add 'Image > Set as Starting Image' menu item

jupdike:
Add 'Set as Starting Image' command with keyboard shortcut, which allows user to click to select an image, then type Cmd+2 to set as starting image without needing to use mouse/context menu. Useful with Cmd+G to add to generation queue using this new starting image.

### DIFF
--- a/Mochi Diffusion/Main Menu/ImageCommands.swift
+++ b/Mochi Diffusion/Main Menu/ImageCommands.swift
@@ -59,6 +59,18 @@ struct ImageCommands: Commands {
             }
             Section {
                 Button {
+                    guard let sdi = store.selected() else { return }
+                    Task { await ImageController.shared.selectStartingImage(sdi: sdi) }
+                } label: {
+                    Text(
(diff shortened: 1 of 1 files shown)


gdbing:
I don't like using ⌘2 for the shortcut, because it's kind of MacOS convention to use ⌘Number to move between UI views or tabs (e.g. Xcode, Mail, Safari, etc)

I don't know what else to use though.

⌘S is the most obvious choice for "Set as starting image" except it is already taken by Save. Mochi doesn't use ⌘I but it is conventionally Get Info, and that is actually a reasonable action for someone to try to take while selecting a gallery image, especially if the right sidebar is collapsed.

I think I'm going to go with ⌘E which is unused, and which could stand for Edit. "Set as starting image" is sort of concepually like editing an image, if you squint a little. And also because it's pretty close to ⌘2, on your keyboard. The only keys which are closer are Q and W, which are both out for obvious reasons.

